PR #1, opened 2020-04-21
bugfix seq calculation if transf. method for CONSERV is set

seb-wahl:
Now namcouple works for the following and we get a 4 in front of flxatmos instead of 3:
```
TotRain OTotRain 16 10800 4 flxatmos EXPOUT
192 96 722 511 atmo opac LAG=0
P 0 P 2
LOCTRANS SCRIPR MAPPING CONSERV
INSTANT
CONSERV D SCALAR LATITUDE 40 FRACNNEI FIRST
rmp_atmo_to_opac_CONSERV_FRACNNEI_D_T63_ORCA05.nc src
GLOBAL
```

### DIFF
--- a/esm_runscripts/oasis.py
+++ b/esm_runscripts/oasis.py
@@ -45,9 +45,8 @@ class oasis:
         # if a transformation method for CONSERV (e.g. GLOBAL) is set below, 
         # increase seq (=number of lines describing the transformation) by 1
         seq = int(direction.get("seq", "2"))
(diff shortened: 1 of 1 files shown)


pgierz:
This is an instance where I'd really like to have CI set up. @dbarbi, maybe one of the new team members can have a look at automatic testing after the release?

`oasis.py` is Dirk's file; and even though the changes look trivial enough, I'd like to be sure: does this still work for AWI-CM?

PR #14, opened 2020-06-24
Oifs stuff

joakimkjellsson:
I've added functionality to use NNOREST, i.e. start OASIS without restart files. This only works for MCT 4.0 and above, so I added mct_version for oasis3mct.yaml. 
I also added an option to turn on the LUCIA analysis tool. 
By default, oasis3mct.yaml sets mct_version = 3.0, NNOREST=F, and use_lucia=False. This should ensure compatibility for those running MCT version < 4.0.


pgierz:
@joakimkjellsson, do you need this in release as well, or can we keep it in develop for a while? I'd like to bundle up a few features before we put everything to release. Also on my list:
* Wind speed errors in ECHAM
* Globbing Fix

Also important: are there any changes needed for the other tools? master, env, config files...?

joakimkjellsson:
No rush at all. I just thought I'd actually commit and pull request the stuff I had before we diverge too much.

seb-wahl:
This has been manually merged into prep_release, see https://github.com/esm-tools/esm_runscripts/pull/30/commits/4e4888a4b16ae8aab82f39837dc5e68af3ef6e23

PR #19, opened 2020-07-27
Hotfix/time step from another model

mandresm:
I defined REcoM's time-step in the `fesom-recom.yaml` as:

```
recom:
    time_step: "${fesom.time_step}"
```

I did this so that the REcoM time step would be always the same as fesom's unless the user defines a specific time step for REcoM.

However, this leads to problems with the calculation of the `prev_date` for REcoM as integers are expected but not variables at the point in `esm_sim_object.py` where `prev_date` is calculated, and `time_step` have not yet been parsed at that point. This merge is to correct that, so that `esm_sim_object.py` can search for the `time_step` at that point.

### DIFF
--- a/esm_runscripts/__init__.py
+++ b/esm_runscripts/__init__.py
@@ -2,7 +2,7 @@
 
 __author__ = """Dirk Barbi"""
 __email__ = 'dirk.barbi@awi.de'
-__version__ = "4.1.4"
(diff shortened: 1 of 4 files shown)


pgierz (commented):
Looks good. I added a docstring and fixed some of the whitespaces.

PR #22, opened 2020-08-07
Hotfix/parent restart dir

mandresm:
`parent_restart_dir` was not well defined on a restart runscript if the model `lresume` was a variable depending on another model.

In the old `esm_sim_objects`:
```
516             if self.config[model]["lresume"] == True and self.config["general"]["run_number"] == 1:
517                 self.config[model]["parent_expid"] = self.config[model][
518                     "ini_parent_exp_id"
519                 ]
520                 if "parent_date" not in self.config[model]:
521                     self.config[model]["parent_date"] = self.config[model][
522                         "ini_parent_date"
523                     ]
524                 self.config[model]["parent_restart_dir"] = self.config[model][
525                     "ini_restart_dir"
526                 ]
527             else:
528                 self.config[model]["parent_expid"] = self.config["general"][
529                     "expid"
530                 ]
531                 if "parent_date" not in self.config[model]:
532                     self.config[model]["parent_date"] = self.config[model][
533                         "prev_date"
534                     ]
535                 self.config[model]["parent_restart_dir"] = self.config[model][
536                     "experiment_restart_in_dir"
537                 ]
```
The following example is for a restart simulation with `echam` and the problems will take place for `jsbach` and `hdmodel`. The first condition `if self.config[model]["lresume"] == True and self.config["general"]["run_number"] == 1:` does not apply for `jsbach` as `self.config[model]["lresume"]`  is not true but it's `'${echam.lresume}'` (that is defined in the `jsbach.inherit.yaml` file). This means it jumps to the `else` for `jsbach` and `hdmodel` and here `parent_restart_dir` is defined as `experiment_restart_in_dir`, no matter what you have defined as `ini_restart_dir`  or `parent_restart_dir`  in the `jsbach.inherit.yaml`.

The issue was solved by adding an if before this block that checks if lresume is a variable and it it is it solves it before entering in the problematic `if`.

Additionally, conditions of the shape `type(object) == type` have been changed to `isinstance(object, type)`.

This is part of a few bug fixes to solve issue https://github.com/esm-tools/esm_tools/issues/77.

### DIFF
--- a/esm_runscripts/__init__.py
+++ b/esm_runscripts/__init__.py
@@ -2,7 +2,7 @@
 
 __author__ = """Dirk Barbi"""
 __email__ = 'dirk.barbi@awi.de'
-__version__ = "4.1.4"
+__version__ = "4.1.5"
 
 from .esm_sim_objects import *
 from .esm_batch_system import *
--- a/esm_runscripts/esm_sim_objects.py
+++ b/esm_runscripts/esm_sim_objects.py
@@ -501,6 +501,9 @@ class SimulationSetup(object):
         for model in self.config["general"]["valid_model_names"]:
             if "time_step" in self.config[model] and not (type(self.config[model]["time_step"]) == str and "${" in self.config[model]["time_step"]):
                 self.config[model]["prev_date"] = self.current_date - (0, 0, 0, 0, 0, int(self.config[model]["time_step"]))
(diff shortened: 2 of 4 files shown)


seb-wahl (approved):
I don't see any problems with your fix.

PR #27, opened 2020-08-23
capitalised Namelist class was not updated in compute.py

mandresm:


pgierz (commented):
Looks good, but note also that the old class name still works:

https://github.com/esm-tools/esm_runscripts/blob/ba42b63703148f6bc75ec011c7b7097840ad4b25/esm_runscripts/namelists.py#L110-L118

My idea was to support both name styles for now, and then remove them once we are sure we have all of them. You can easily turn warnings into errors so we can find out where the old names are still being used.

However, there's a mistake in that snippet. It should of course be `class namelist` instead of `def namelist`

PR #29, opened 2020-08-28
added a better exception handling to jobclass.py -> find_correct_source

denizural:
Sometimes the **forcing_source** is a dictionary and esm_runscript ends with a cryptic error like:

>   File "/sw/rhel6-x64/conda/anaconda3-bleeding_edge/lib/python3.6/posixpath.py", line 146, in basename
>     p = os.fspath(p)
> TypeError: expected str, bytes or os.PathLike object, not dict

I added a more explanatory error message with an Exception so that the user can understand what went wrong. So, it now does this:

> ==============================================
>   ERROR: file_source is not of type string or Path.
>          Exiting the program
>   - file_source:
>  {'/pool/data/ECHAM6/input/r0007/T127/aero/T127_aeropt_kinne_sw_b14_fin_1865.nc': {'to': 1864}, '/pool/data/ECHAM6/input/r0007/T127/aero/T127_aeropt_kinne_sw_b14_fin_2000.nc': {'from': 1865, 'to': 2000}}
> 
>   - type(file_source) : <class 'dict'>
> ==============================================
> 
> Traceback (most recent call last):
>   File "/pf/a/a271096/.local/bin/esm_runscripts", line 11, in <module>
>     load_entry_point('esm-runscripts', 'console_scripts', 'esm_runscripts')()
>   File "/mnt/lustre01/pf/a/a271096/esm_all_packages/esm_runscripts/esm_runscripts/cli.py", line 168, in main
>     Setup()
>   File "/mnt/lustre01/pf/a/a271096/esm_all_packages/esm_runscripts/esm_runscripts/esm_sim_objects.py", line 54, in __call__
>     self.compute(*args, **kwargs)
>   File "/mnt/lustre01/pf/a/a271096/esm_all_packages/esm_runscripts/esm_runscripts/esm_sim_objects.py", line 79, in compute
>     Compute = compute(self.config)
>   File "/mnt/lustre01/pf/a/a271096/esm_all_packages/esm_runscripts/esm_runscripts/compute.py", line 13, in __init__
>     self.all_files_to_copy = self.assemble_file_lists(config, self.relevant_files)
>   File "/mnt/lustre01/pf/a/a271096/esm_all_packages/esm_runscripts/esm_runscripts/jobclass.py", line 40, in assemble_file_lists
>     self.really_assemble_file_list(config, model, filetypes)
>   File "/mnt/lustre01/pf/a/a271096/esm_all_packages/esm_runscripts/esm_runscripts/jobclass.py", line 244, in really_assemble_file_list
>     source_name=self.find_correct_source(file_source, year)
>   File "/mnt/lustre01/pf/a/a271096/esm_all_packages/esm_runscripts/esm_runscripts/jobclass.py", line 314, in find_correct_source
>     raise TypeError("file_source does not have the correct type")
> TypeError: file_source does not have the correct type

### DIFF
--- a/esm_runscripts/jobclass.py
+++ b/esm_runscripts/jobclass.py
@@ -305,6 +305,23 @@ class jobclass:
                     return fname
                 else:
                     continue
+                
+            # Deniz: added error checking. Sometimes file_source becomes a dictionary
+            # instead of a string.
+            import pathlib
+            if not isinstance(file_source, str) or not instance(file_source, pathlib.Path):
+                print("==============================================")
+                print("  ERROR: file_source is not of type string or Path.")
+                print("         Exiting the program")
+                print("  - file_source: \n {}".format(file_source))
+                print()
+                print("  - type(file_source) : {}".format(type(file_source)))
+                print("==============================================")
+                raise TypeError("file_source does not have the correct type")
+
+                #import sys
+                #sys.exit(-1)
+
         return file_source
 
 


pgierz (commented):
Looks good. Were you able to figure out why the fie_source is a dict sometimes, and sometimes not? That'd be good to know...

Maybe it is an idea to just directly do a sys.exit rather than raising the error. I think one of the things @seb-wahl has always been asking for is "clean errors with nice exits", so, I'd say if we catch an understood error, we just exit right away rather than raising an error again. 

As a different idea, we could raise tracebacks and errors if there is something like "debug_mode" in the general section of the config.

pgierz (changes requested):
Also, put the import at the top of the file...

denizural:
**1) Error message:**
Actually, I first started with the sys.exit(1). As you can see it is commented. The reason for raising a TypeError is that first, I make an error message (I hope @seb-wahl finds that clean :-) ), and then print out the whole error stack so that we know where it got messy. 

When I first exitted with the error message, I was somehow not happy about it since I want to go to the exact point with a debugger and for that having an error stack is good. But an end user may find it frightening.

**2) Dictionary stuff:**
Suppose that you use **histaerofin** and echam.datasets.yaml specifies that

> "histaerofin":
>         "${forcing_dir}/aero/${resolution}_aeropt_kinne_sw_b14_fin_1865.nc":
>                 to: 1864
>         "${forcing_dir}/aero/${resolution}_aeropt_kinne_sw_b14_fin_@YEAR@.nc":
>                 from: 1865
>                 to: 2000

If the current year is 2001 then basically it returns the dictionary

{'/pool/data/ECHAM6/input/r0007/T127/aero/T127_aeropt_kinne_sw_b14_fin_1865.nc': {'to': 1864}, '/pool/data/ECHAM6/input/r0007/T127/aero/T127_aeropt_kinne_sw_b14_fin_@YEAR@.nc': {'from': 1865, 'to': 2000}}

since there is no check for that. Of course this is an user error. They should know what dataset to choose and its time range but they end with a cryptic error at the end. 

As I mentioned before in the group meeting, checking the **return type** is important. Python usually does a good job but as programmers we need to make sure that it does not break anything that depends on it.

denizural:
> Also, put the import at the top of the file...

Yes, I know. Just to keep things compact. Normally I also add them to top. 

It is important to note that some esm functions don't have "import logging" while they use logging.debug(). logging module is imported in the top of the file but Python gives `logging is not defined` error.

pgierz:
> logging module is imported in the top of the file but Python gives logging is not defined error.

That sounds really strange....do you have an example of that?

mandresm (approved):
I think everything is fine, but also agree that @pgierz's suggestions need to be incorporated.

About the cleaner error messages, I agree with @seb-wahl, but maybe this is something that should be addressed for the whole code (and not only for this particular error) under a common strategy:
- Should we create or own exception classes for errors or just use `if`s as used here?
- How much error information should we give?
- What should the debug mode return?

I think this is something we need to talk about in a meeting and agree on a common strategy, and then make the required changes. Therefore, my opinion is that this changes need to be implemented as they are (correcting the position of the `import`s) and we can later address a common strategy for error handling.

seb-wahl:
> > Also, put the import at the top of the file...
> 
> Yes, I know. Just to keep things compact. Normally I also add them to top.
> 
> It is important to note that some esm functions don't have "import logging" while they use logging.debug(). logging module is imported in the top of the file but Python gives `logging is not defined` error.

Another argument working on https://github.com/esm-tools/esm_tools/issues/116 as soon as possible :-)

dbarbi:
Not needed anymore as the whole function was thrown away

PR #30, opened 2020-08-31
enable correct namcouple generation for oasis3mct version 4.0

seb-wahl:
manually added fixes for OASIS-MCT 4.0 from branch `oifs_stuff`. Use get instead of ['...'] in some places to avoid dependency on latest version of esm_tools / oasis3mct.yaml
tested with oasis as part of FOCI and FOCI-OIFS.


pgierz (commented):
I'm not sure I like using floats as version numbers. Maybe we can pass them in as strings, and then convert them to tuples? 

```python
version = "4.0"
version = tuple(int(x) for x in version.split("."))
# A comparison:
version > (2, 8)  # Gives True
```

PR #37, opened 2020-09-15
Feature/cleanup run dir

pgierz:
# New Feature

This allows you to cleanup the `run_????` directories. The user has the following options available in their runscript:

```yaml
general:
    remove_this_rundir: True  # throws away the rundir that was just finished
    remove_old_rundirs_except: 5 # Throws away everything except the last 5 run dirs
   remove_old_rundirs_keep_every: 2 # throw away everything except every 2nd run dir
   remove_log_dir: True # Also valid for any other file type. Throws away the log dir of this run
   remove_size: 37890 # Throws away any file greater than x bytes (sorry, needs to be an int)
```

# TODO:
- [x] Actual implementation
- [x] Test drive it (@mandresm, can you see if this even works?)
- [x] User docs: should be in the handbook somewhere. There might be a way to directly include the doc-string from the main function as the user how-to 
- [x] Code docs (sort of started, but I could probably write more docstrings)

# Related Issues:
Will Close https://github.com/esm-tools/esm_tools/issues/96, https://github.com/esm-tools/esm_tools/issues/55

### DIFF
--- a/esm_runscripts/esm_sim_objects.py
+++ b/esm_runscripts/esm_sim_objects.py
@@ -835,24 +835,78 @@ class SimulationSetup(object):
     # @staticmethod
     def clean_run_dir(self, config):
         """
-        This plugin allows you to clean up the ``run_${DATE}`` folder. The
-        cleanup files are specified in the config ``general`` section as
-        follows (documentation follows order of code as it is executed):
+        This plugin allows you to clean up the ``run_${DATE}`` folders.
+        To do that you can use the following variables under the
+        ``general`` section of your runscript (documentation follows order
+        of code as it is executed):
 
-        * ``general.clean_this_rundir``: (bool) Removes the entire run
-          directory.
+        * ``clean_runs``: **This is the most important variable for most
+          users**. It can take the following values:
+            * ``True``: removes the ``run_`` directory after each run
+              (**overrides every other** ``clean_`` **option**).
 
-        * ``general.clean_old_rundirs_except``: (int) Removes the entire run
-          directory except for the last <x> runs
+            * ``False``: does not remove any ``run_`` directory (default)
+              if no ``clean_`` variable is defined.
 
-        * ``general.clean_old_rundirs_keep_every``: (int) Removes the entire
-          run directory except every <x>th run
+            * ``<int>``: giving an integer as a value results in deleting
+              the ``run_`` folders except for the last <int> runs
+              (recommended option as it allows for debugging of crashed
+              simulations).
 
-        * ``general.clean_<filetype>_dir``: (bool) Erases the run directory
-          for a specific filetype
+          .. Note::
+             ``clean_runs: (bool)`` is incompatible with
+             ``clean_this_rundir`` and ``clean_runs: (int)`` is incompatible
+             with ``clean_old_rundirs_except`` (an error will be raised
+             after the end of the first simulation). The functionality of
+             ``clean_runs`` variable **alone will suffice most of the
+             standard user requirements**. If finer tunning for the removal
+             of ``run_`` directories is required you can used the following
+             variables instead of ``clean_runs``.
 
-        * ``general.clean_size``: (int or float) Erases all files with size
-          greater than ``clean_size``, must be specified in bytes!
+        * ``clean_this_rundir``: (bool) Removes the entire run directory
+          (equivalent to ``clean_runs: (bool)``). ``clean_this_rundir: True``
+          **overrides every other** ``clean_`` **option**.
+
+        * ``clean_old_rundirs_except``: (int) Removes the entire run
+          directory except for the last <x> runs (equivalent to
+          ``clean_runs: (int)``).
+
+        * ``clean_old_rundirs_keep_every``: (int) Removes the entire
+          run directory except every <x>th run. Compatible with
+          ``clean_old_rundirs_except`` or ``clean_runs: (int)``.
+
+        * ``clean_<filetype>_dir``: (bool) Erases the run directory
+          for a specific filetype. Compatible with all the other options.
+
+        * ``clean_size``: (int or float) Erases all files with size
+          greater than ``clean_size``, must be specified in bytes! Compatible
+          with all the other options.
+
+        Example
+        -------
+
+        To delete all the ``run_`` directories in your experiment include this
+        into your runscript:
+
+        .. code-block:: yaml
+
+           general:
+                   clean_runs: True
+
+        To keep the last 2 ``run_`` directories:
+
+        .. code-block:: yaml
+
+           general:
+                   clean_runs: 2
+
+        To keep the last 2 runs and every 5 runs:
+
+        .. code-block:: yaml
+
+           general:
+                   clean_old_rundirs_except: 2
+                   clean_old_rundirs_keep_every: 5
         """
         self._clean_run_determine_user_choice(config)
         self._clean_this_rundir(config)
--- a/esm_runscripts/esm_sim_objects.py
+++ b/esm_runscripts/esm_sim_objects.py
@@ -5,6 +5,7 @@ from datetime import datetime
 from io import StringIO
 import collections
 import logging
+import pathlib
 import os
 import pdb
 import shutil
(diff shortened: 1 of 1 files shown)


seb-wahl (changes requested):
```yaml
general:
    remove_this_rundir: True  # throws away the rundir that was just finished
    remove_old_rundirs_except: 5 # Throws away everything except the last 5 run dirs
   remove_old_rundirs_keep_every: 2 # throw away everything except every 2nd run dir
   remove_log_dir: True # Also valid for any other file type. Throws away the log dir of this run
   remove_size: 37890 # Throws away any file greater than x bytes (sorry, needs to be an int)
```
I couldn't spot any error in the implementation but I would keep it simpler and add just one option:
`remove_rundirs : True or integer >=0` with the following meaning
* True: same as `remove_this_rundir: True` in the current implementation
* integer > 0: keep last x rundirs
* False or 0: current behaviour (keep all rundirs)

Also I didn't get what `remove_log_dir` should do. Which log dir do you mean? I think logs of a run should always be kept. Rather we should offer an option to `tar.gz` the logs during postprocessing.
And I couldn't think of a use case for the `remove_size` option.

pgierz:
The `remove_log_dir` was just an example. You could also put, for instance, `remove_restart_dir`, or whatever else you want. This refers to the `filetype` dir specifically in the `run_???` folder.

I can certainly make it easier with just one option, but that makes it a bit less flexible with regards to keeping specific runs, such as every 10th or so on. That might be valuable for investigating errors after the fact

pgierz:
@seb-wahl I just wrote a bit that simplifies the options, but the other, more granular things are still permitted. Let me know what you think.

seb-wahl (approved):
I'm happy now :-) I'll test it once I'm done with other requests that are on top of my to do list. Probably tomorrow morning.

seb-wahl:
> I'm happy now :-) I'll test it once I'm done with other requests that are on top of my to do list. Probably tomorrow morning.

Doesn't work for now (or is it me doing something wrong), runscript is `runscripts/oifs/oifs-43r3-tco95-amip-blogin-initial-daily_SW.yaml"` from branch `feature/add-oifs-43r3-v1` (being on the oifs branch shouldn't matter, right?):

```yaml
general:
        setup_name: oifs
        compute_time: "00:30:00"
        account: shkifmsw
        initial_date: "2008-01-01"
        final_date: "2008-01-03"
        model_dir: /home/shkifmsw/esm/models/oifs-43r3-v1/
        base_dir: /home/shkifmsw/esm/esm-experiments/
        nyear: 0
        nmonth: 0
        nday: 1
        restart_rate: 0
        restart_unit: days
        # cleanup of rundir
        remove_runs: True

oifs:
....
....
```

```bash
blogin3:~/esm/esm_tools/runscripts/oifs $ esm_runscripts -e oifs-xios-4 oifs-43r3-tco95-amip-blogin-initial-daily_SW.yaml -c
Started from:  /home/shkifmsw/esm/esm_tools/runscripts/oifs
starting :  compute
Traceback (most recent call last):
  File "/home/shkifmsw/.local/bin/esm_runscripts", line 11, in <module>
  File "/home/shkifmsw/gitlab/esm_runscripts/esm_runscripts/cli.py", line 167, in main
    Setup = SimulationSetup(command_line_config)
  File "/home/shkifmsw/gitlab/esm_runscripts/esm_runscripts/esm_sim_objects.py", line 85, in __init__
    self.get_total_config_from_user_config(user_config)
  File "/home/shkifmsw/gitlab/esm_runscripts/esm_runscripts/esm_sim_objects.py", line 278, in get_total_config_from_user_config
    esm_parser.choose_blocks(self.config, blackdict=self.config._blackdict)
  File "/home/shkifmsw/.local/lib/python3.7/site-packages/esm_parser/esm_parser.py", line 2071, in choose_blocks
    remove_entries_from_chapter_in_config(config, all_names, config, all_names)
  File "/home/shkifmsw/.local/lib/python3.7/site-packages/esm_parser/esm_parser.py", line 590, in remove_entries_from_chapter_in_config
    setup_config,
  File "/home/shkifmsw/.local/lib/python3.7/site-packages/esm_parser/esm_parser.py", line 542, in remove_entry_from_chapter
    raise TypeError("Please put all entries to remove as a list")
TypeError: Please put all entries to remove as a list
blogin3:~/esm/esm_tools/runscripts/oifs $ esm_versions check
You are using the following esm_tools versions:
-----------------------------------------------
esm_archiving : unknown version!
esm_autotests : unknown version!
esm_calendar: 4.0.1
esm_database: 4.0.0
esm_environment: 4.2.0
esm_master: 4.2.2
esm_parser: 4.2.1
esm_profile: 4.0.0
esm_rcfile: 4.0.0
esm_runscripts: 4.2.1 (development install, on branch: feature/cleanup_run_dir, describe=v4.2.1-3-g7bcbcd8)
esm_tools: 4.2.0 (development install, on branch: feature/add-oifs-43r3-v1, describe=v4.2.0-5-g82e2abe-dirty)
esm_plugin_manager: 4.2.0
esm_version_checker: 4.4.0
```
***Ideas?***

pgierz:
OK, super super dumb, but we can't use the word "remove_entries", since the parser uses it to determine what to add and remove from other chapters. Perhaps we can change it to "clean"?

pgierz:
@seb-wahl, can you try again after you pull, but change your general.remove_runs to general.clean_runs?

esm_runscripts/esm_sim_objects.py
```diff
@@ -710,10 +750,167 @@ def tidy(self):
...
 
+        * ``True`` Removes the current run dir
+        * ``False`` Keeps run dir
+        * ``int`` (must be >= 0) keep last ``x`` run dirs
+        """
+        user_clean = config["general"].get("clean_runs")
+        # TODO(PG): It might be nice if these sorts of checks happened earlier
+        # in the job, before it even gets to this function
+        if isinstance(user_clean, bool):
+            if "clean_this_run" not in config["general"]:
```
mandresm: I think this `clean_this_run` and later occurences should be changed to `clean_this_rundir` though this did not fix the error.
```
501 ------------------------------------------
502 Type Error!
503 You have set this in your config:
504 general:
505     clean_runs:  None
506 
507 This is of type:  <class 'NoneType'>
508 However, only the following types are valid:
509    * boolean
510    * integer (greater or equal to 0!)
511 Please correct that
512 ------------------------------------------
```

pgierz: Yes, also the entire thing needs to be wrapped in a `if user_clean:`. I'll push a commit in a second, if you can try again that'd be great!

mandresm: That make sense. Then the user will have 2 options?

- Option 1: simple syntax
`clean_runs` to be define as `True`, to delete all run folders, `False` to delete nothing, and a number to preserver X number of runs before the current date
- Option 2: more complex syntax where you can define the following variables: `clean_this_rundir`, `clean_old_rundirs_except`, `clean_old_rundirs_keep_every`, `clean_<filetype>_dir`, `clean_size`

pgierz: Yup exactly. The easy one should suffice for most people, but it might be that you need more granular control for whatever reason.

mandresm:
The option `clean_old_rundirs_keep_every` was not working because `all_run_folders_in_experiment` contained only the existing folders which means the index operation `all_run_folders_in_experiment[::number_rundirs_keep_every]` won't do what was expected. I fixed it by making a new class `RunFolders` and making `all_run_folders_in_experiment` be an instance of it. This instance preserves the names of the `run_` folders even if they are deleted, enabling the index operations we want to perform.

The `_except` option still has a problem. I'll keep having a look at it.

Another question: I think the file `<exp_id>_compute_<date1>-<date2>_<JobID>.log>` in `<exp_id>/run_<date>/scripts` is very valuable but at the moment is not transferred by defect to the general experiment folder, which means, it gets deleted when using this new functionality. Shouldn't we do something so that it is preserved (e.g. copy it to `<exp_id>/scripts`)?

esm_runscripts/esm_sim_objects.py
```diff
@@ -70,6 +70,86 @@ def size_human_to_bytes(s, suffix="B"):
         num *= 1024.0
     return num
 
+class RunFolders(list):
```
pgierz, line 73: This is actually really really cool 👍

pgierz:
> Another question: I think the file <exp_id>_compute_<date1>-<date2>_<JobID>.log> in <exp_id>/run_<date>/scripts is very valuable but at the moment is not transferred by defect to the general experiment folder, which means, it gets deleted when using this new functionality. Shouldn't we do something so that it is preserved (e.g. copy it to <exp_id>/scripts)?

Maybe that should be written directly to `<exp_id>/scripts`, or put there as part of the tidy job before we nuke the `run_???` folders

esm_runscripts/esm_sim_objects.py
mandresm: Thanks :)

mandresm:
> Maybe that should be written directly to `<exp_id>/scripts`, or put there as part of the tidy job before we nuke the `run_???` folders

I think you are right with the first one, in this way we can keep logging even after cleaning.

mandresm:
Tests progress:

- [x] `clean_runs: True`
- [x] `clean_runs: False`
- [x] `clean_runs: <int>`
- [x] `clean_this_rundir`
- [x] `clean_old_rundirs_except`
- [x] `clean_old_rundirs_keep_every`
- [x] `clean_old_rundirs_except` with `clean_old_rundirs_keep_every`
- [x] `clean_<type>_dir`
- [x] `clean_size`
- [x] test errors for `clean_runs` with `clean_this_rundir` and/or `clean_old_rundirs_except`

mandresm:
I finally finished with all the testing and fixing some bugs. The behavior is as expected with all the combinations I could think of. There is only the following thing:

- [x] `clean_runs: True` and `clean_this_rundir` override all the other clean options as they are run before. I'm fine with this, we just need to make sure we **include it in the user manual**.

mandresm:
>     * User docs: should be in the handbook somewhere. There might be a way to directly include the doc-string from the main function as the user how-to

There is infact a way: using `.. automethod:: esm_runscripts.SimulationSetup.clean_run_dir` inserts the docstrings in the `.rst`.

Here some documentation about it: https://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html

mandresm (approved):

PR #38, opened 2020-10-09
[FEATURE] no more file.log

pgierz:
Removes the annoying `file.log` that was generated whenever you used `esm_runscripts <config> <opts>`. It was always empty and never contained any info anyway...


pgierz:
will be in prep release branch

PR #39, opened 2020-10-19
[FEATURE] PISM Support

pgierz:
Support for the PISM ice sheet model.

### DIFF
--- a/esm_runscripts/compute.py
+++ b/esm_runscripts/compute.py
@@ -19,9 +19,6 @@ from .namelists import Namelist
 
 class compute(jobclass):
     def __init__(self, config):
-
-        self.relevant_files = ["bin", "config", "forcing", "input", "restart_in"]
-        self.all_files_to_copy = self.assemble_file_lists(config, self.relevant_files)
         # Check for a user defined compute recipe in the setup section of the
         # general section. If nothing is found, recipe_steps should evaluate to
         # None and the default is used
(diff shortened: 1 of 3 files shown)


esm_runscripts/compute.py
```diff
@@ -19,21 +19,48 @@
...
+            The entire experiment configuration
+
+        """
+        self = config["general"]["jobclass"]
+        relevant_files = ["bin", "config", "forcing", "input", "restart_in"]
+        all_files_to_copy = self.assemble_file_lists(config, relevant_files)
+        config["general"]["all_files_to_copy"] = all_files_to_copy
+        # FIXME: This should always just come from the config and not from self:
+        self.all_files_to_copy = all_files_to_copy
+        import pdb; pdb.set_trace()
```
pgierz: this needs to be removed

esm_runscripts/compute.py
```diff
@@ -19,21 +19,48 @@
...
+
+        Returns
+        -------
+        config : dict
+            The entire experiment configuration
+
+        """
+        self = config["general"]["jobclass"]
+        relevant_files = ["bin", "config", "forcing", "input", "restart_in"]
+        all_files_to_copy = self.assemble_file_lists(config, relevant_files)
```
pgierz, line 57: this method could use a doc-string so we know what it does. it sends back a list, apparently

esm_runscripts/compute.py
```diff
@@ -19,21 +19,48 @@
...
+            The entire experiment configuration
+
+        Returns
+        -------
+        config : dict
+            The entire experiment configuration
+
+        """
+        self = config["general"]["jobclass"]
+        relevant_files = ["bin", "config", "forcing", "input", "restart_in"]
```
pgierz, line 56: Check if any other methods refer to `self.relevant_files`. As it is now, this is just a local variable

esm_runscripts/compute.py
```diff
@@ -19,21 +19,48 @@
...
+        config : dict
+            The entire experiment configuration
+
+        Returns
+        -------
+        config : dict
+            The entire experiment configuration
+
+        """
+        self = config["general"]["jobclass"]
```
pgierz, line 55: This sort-of defeats the purpose of having static methods...

PR #40, opened 2020-10-20
[FEATURE] Logging overhaul

pgierz:
Overall of logging, as requested by @seb-wahl in https://github.com/esm-tools/esm_tools/issues/116.

Configuration over the runscript is not yet possible, but you can specify everything via environmental variables. Please see: https://loguru.readthedocs.io/en/stable/api/logger.html#env

Here's what it looks like out-of-the box:

Screen Shot 2020-10-20 at 1 53 22 PM

Closes esm-tools/esm_tools#116.

### DIFF
--- a/esm_runscripts/compute.py
+++ b/esm_runscripts/compute.py
@@ -5,6 +5,8 @@ import os
 import shutil
 import sys
 
+from loguru import logger
+
 import esm_rcfile
 import six
 import yaml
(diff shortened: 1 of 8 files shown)


seb-wahl (approved):
Looks fine to me. Before this goes into release we need to make sure that a `esm_versions upgrade` correctly installs the new dependency (loguru) to avoid problems for the users.

pgierz:
I still want to find a way to get the configuration possible from the user runscript as well as over environmental variables before I change it from draft to actual pull request, but I probably won't have time until end of the week.

PR #41, opened 2020-10-22
Update runscripts in the experiment directory

mandresm:
This hotfix solves issue esm-tools/esm_tools#175 in the following way:
- If running `esm_runscript` without `-U`, and `${base_dir}/${experiment_id}/scripts/<runscript_invoked_by_the_user>.yaml`  exists then reports an **error** saying that the runscript already exists and that the user needs to use the `-U` flag which means that the runscript will be overwritten. 
- This error only happens if the runscripts are not identical. If they are not identical the error returns the difference in a similar format as diff .
- If running `esm_runscript` with `-U` , and `${base_dir}/${experiment_id}/scripts/<runscript_invoked_by_the_user>.yaml` exists, overwrites it. Differences are also reported back to the user.

### DIFF
--- a/esm_runscripts/compute.py
+++ b/esm_runscripts/compute.py
@@ -244,46 +244,69 @@ class compute(jobclass):
 
     @staticmethod
     def copy_tools_to_thisrun(config):
+        """
+        Copies the tools, namelists and runscripts to the experiment directory,
+        making sure that they don't overwrite previously existing files unless
+        the ``-U`` flag is used.
 
+        Parameters
+        ----------
+        config : dict
+            Dictionary containing the configuration information.
+        """
         gconfig = config["general"]
 
+        # Directory where the original runscript is
         fromdir = os.path.realpath(gconfig["started_from"])
+        # Directory where the runscript is copied to
         scriptsdir = os.path.realpath(gconfig["experiment_scripts_dir"])
 
+        # Paths inside the experiment directory where esm_tools and namelists
+        # are copied to. Those are not functional but a reference to what was
+        # the original state when the experiment was firstly started
         tools_dir = scriptsdir + "/esm_tools/functions"
         namelists_dir = scriptsdir + "/esm_tools/namelists"
 
         print("Started from :", fromdir)
         print("Scripts Dir : ", scriptsdir)
 
+        # Update namelists and esm_tools. These have no effect on the final
+        # simulation as only the installed esm_tools with their runscripts
+        # are used, unless esm_tools and namelists paths are manually defined
+        # by the model or user (not recommended for the esm_tools path)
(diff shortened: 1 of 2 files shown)


mandresm:
@denizural , @dbarbi , @pgierz , I modify the behavior of this feature to include a question for the user. Now it does the following:

- If running `esm_runscript` without `-U`, and `${base_dir}/${experiment_id}/scripts/<runscript_invoked_by_the_user>.yaml`  exists then:
    - asks if it should update the runscript in the experiment with the new one. If yes, updates it, if not reports an **error**.
- The question is only posed if the runscripts are not identical. 
- If running `esm_runscript` with `-U` , and `${base_dir}/${experiment_id}/scripts/<runscript_invoked_by_the_user>.yaml` exists, overwrites it. Differences are also reported back to the user.

esm_runscripts/compute.py
```diff
@@ -298,6 +325,112 @@ def copy_tools_to_thisrun(config):
             gconfig["profile"] = False
             compute.end_it_all(config, silent=True)
 
+    @classmethod
```
pgierz: why is this a class method? I don't see the `cls` been used anywhere. Can it just be a function instead?

esm_runscripts/compute.py
```diff
@@ -244,46 +244,73 @@ def _write_finalized_config(config):
 
     @staticmethod
     def copy_tools_to_thisrun(config):
```
pgierz, line 246: One thing I never understood about this method: We copy everything to the experiment. That is nice. However, I think the actual yaml files and namelists are still read from wherever the `.esmtoolsrc` file points to?

esm_runscripts/compute.py
mandresm: You are right, there is no need of having that done with a `@classmethod`. I had some issues calling the this method at the beginning from a `@staticmethod` but I was just missing to refer to the `compute` instanced. When I remove `@classmethod` and the input variable `cls` it still works as intended, so I will do that.

esm_runscripts/compute.py
mandresm: Yes, this thing was doing virtually nothing, because it would only update the `namelists` and `esm_tools` when using  `-U` but not the runscript, and it is only the runscript that is actually used. That is why I include the runscript. I left the rest as it was cause I thought there should be a reason for that, but if it is a deprecated feature maybe we can think of getting rid of the `namelist` and `esm_tools` copying.

mandresm:
This pull request should not be merged!

dbarbi:
Merged this into paul_sandbox by hand, and will close the request now. Can I remove the branch?

mandresm:
Sure, go ahead

PR #43, opened 2020-10-28
removed export_mode from oasis.py. Instead set by full_config[self.na…

joakimkjellsson:
I realised there's no way to control "export_mode" for OASIS in ESM-Tools. 
I would like the user to be able to set "EXPORTED" or "EXPOUT" (write coupling fields to disk) in oasis3mct.yaml

The old behaviour was to use EXPOUT for initial runs and EXPORTED when resume=True. 
I've changed so that the user sets export_mode in oasis3mct.yaml, similar to how coupling_frequency is set. 
If nothing is set, the model defaults to EXPORTED. 

Does this sound like a a good addition to ESM-Runscripts? 

Cheers
Joakim

### DIFF
--- a/esm_runscripts/esm_coupler.py
+++ b/esm_runscripts/esm_coupler.py
@@ -230,8 +230,12 @@ class esm_coupler:
                         if "coupling_methods" in full_config[self.name]:
                             if interpolation in full_config[self.name]["coupling_methods"]:
                                 transf_info=full_config[self.name]["coupling_methods"][interpolation]
-
-                        self.coupler.add_coupling(lefts, lgrid_info, rights, rgrid_info, direction_info, transf_info, restart_file, full_config[self.name]["coupling_time_step"], full_config[self.name]["lresume"])
+                        if "export_mode" in full_config[self.name]:
+                            export_mode = full_config[self.name]["export_mode"]
+                        else:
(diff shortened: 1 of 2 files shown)


pgierz:
What would you put in the user runscript? Like this?

```yaml
oasis3mct:
   export_mode: EXPORTED
```

joakimkjellsson:
Yep. That would do it. 
Or 
```yaml
choose_resume: 
          1: 
                   export_mode: EXPORTED
          0:
                   export_mode: EXPOUT
```
in  e.g. config/setups/focioifs/focioifs.yaml to recover the old behaviour.

dbarbi:
ok, merged this by hand. I made a change though - the default value should not be EXPORTED, but EXPOUT in the first year. so I changed the default to DEFAULT and added this:

```
if "export_mode" in full_config[self.name]:
                            export_mode = full_config[self.name]["export_mode"]
                        else:
                            export_mode = "EXPORTED" 
                        print("DEBUG: EXPORT_MODE: ",export_mode)
                        self.coupler.add_coupling(lefts, lgrid_info, rights, rgrid_info, direction_info, transf_info, restart_file, full_config[self.name]["coupling_time_step"], full_config[self.name]["lresume"], export_mode=export_mode)
```

Can I remove this branch?

PR #44, opened 2020-10-29
ulimit

pgierz:
Sets stack size to unlimited before the run


pgierz:
Will be in prep release branch

PR #50, opened 2020-12-08
Bugfix: Added missing yac.py

nwieters:
I noticed that this is still missing in esm_runscripts.


nwieters:
I also needed to fix the coupling file handling to include coupling.xml file for YAC coupler.

dbarbi (approved):

PR #53, opened 2020-12-11
Exception to add_<env_vars> added to the environment case checker

mandresm:
Exceptions for export_vars and module actions were already present for the new environment case checker (#48), but an error could potentially occur when there was a add_export_vars or add_module_actions in the environment dictionaries of a component. add_export_vars and add_module_actions are added here as exceptions for the environment case checker.

Solves esm-tools/esm_tools#248


seb-wahl:
I can't judge whether this breaks any of my setups, but I don't think so.

PR #56, opened 2020-12-17
hotfix copy namelists

mandresm:
Fixed an issue for which only namelist containing the string namelist would be copied.

Solves esm-tools/esm_tools#255

### DIFF
--- a/esm_runscripts/filelists.py
+++ b/esm_runscripts/filelists.py
@@ -558,10 +558,11 @@ def copy_files(config, filetypes, source, target):
                     # namelists that have no ``namelist`` in their name. This is a more
                     # general implementation but it enforces the use of the
                     # ``namelists`` list to be defined for each model with namelists.
-                    isnamelist = (
-                        "namelist" in file_source or
-                        any(x in file_source for x in config[model].get("namelists", []))
+                    namelist_candidates = (
+                        [item for item in config[model].get("namelists", [])]
+                        + ["namelist"]
                     )
+                    isnamelist = any(map(file_source.__contains__, namelist_candidates))
                     if source == "init":
                         if isnamelist and file_source.startswith("NONE_YET"):
                             file_source = esm_tools.get_namelist_filepath(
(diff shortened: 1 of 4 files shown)


esm_runscripts/filelists.py
```diff
@@ -554,10 +554,16 @@ def copy_files(config, filetypes, source, target):
...
                     # NOTE(PG): This is a really, really, REALLY bad hack and it
                     # makes me physically ill to look at:
+                    # NOTE(MA): The previous implementation was not able to include
+                    # namelists that have no ``namelist`` in their name. This is a more
+                    # general implementation but it enforces the use of the
+                    # ``namelists`` list to be defined for each model with namelists.
+                    isnamelist = (
+                        "namelist" in file_source or
+                        any(x in file_source for x in config[model].get("namelists", []))
+                    )
```
denizural, line 564: It looks to be working since I tried it in manually with Fesom namelists. However, I would write the checking bit differently:

```
# list of strings that contains possible namelist files
namelist_candidates = [item for item in config['model'].get("namelists", [])] + ['namelist']

# 1st way: use cool functional programming to search the search for namelist candiates in the file_source
any(map(file_source.__contains__, namelist_candidates))

# 2nd way: or use list comprehension
any(string in file_source for string in namelist_candidates)
```

So, first I created a list of possible namelist `candidates`. Then, I use either very clean map expression (1st way) or a simpler generator expression than on line 563. I thinks this is better of readability and more 'Pythonic'.

denizural (approved):

esm_runscripts/filelists.py
mandresm: You're right, by why to use a list comprehension in this suggested step?
```
# list of strings that contains possible namelist files
namelist_candidates = [item for item in config['model'].get("namelists", [])] + ['namelist']
```
Couldn't we just use?
```
namelist_candidates = config['model'].get("namelists", []) + ['namelist']
```

denizural: Oh yes, it looks even better!

mandresm: Some times `config['model']["namelists",]` is a dictionary and other times it is a list (depending on if the namelists have been already loaded or not), therefore there is the need to use the list comprehension as you suggested.

Thanks the review :)

PR #61, opened 2021-01-04
fix the ulimit -s

denizural:
`ulimit -s` is a get command it does not set any value. As you can see in the log file it prints the stack size (usually `unlimited`).

Instead, it should be `ulimit -s unlimited` which actually sets the stack size.


pgierz (approved):

pgierz:
Looks good Deniz. Go ahead and merge, and remember to bump the version numbers and give a new tag on the release branch. 

I'll look into setting up some sort of automation for that, I think GitHub actions can auto-bump for us, but I need to still make sure it works correctly...

PR #62, opened 2021-01-11
Fifex wrong indentation of elif self.coupler.name == "yac": (line 250…

nwieters:
Fixed wrong indentation of elif self.coupler.name == "yac": (line 250). Applied flake8 suggestions.


dbarbi (approved):
Sure. Will merge this by hand now.

PR #69, opened 2021-01-25
subdirs on targets

mandresm:
It was not possible for `esm_runscripts` to copy files into a subfolder of the main experiment folders when they were coming directly from the source, and not from the respective subfolder in the source (i.e. copy a file from `work exp_id/run_DATE/work/a_file` into a subfolder in outdata `exp_id/outdata/model/subfolder/a_file`). This hotfix solves the issue.


seb-wahl (approved):
Didn't test it, but I couldn't spot an error. Fingers crossed.

PR #74, opened 2021-02-03
FEATURE: Multi Srun

pgierz:
This one is going to probably require a video chat to fully explain

For part of PalMod, we want to be able to run 2 (or more) models in parallel within the same SLURM job. To do that, we inject a few things into the sad file:

* Figure out which nodes we are working on
* Assign specific nodes to each of the models via the srun command `--nodelist`

Note that this whole thing only actually starts if you have a "multi_srun" part in your model. I **hope** that the normal cases won't be affected but please check.

### DIFF
--- a/esm_runscripts/batch_system.py
+++ b/esm_runscripts/batch_system.py
@@ -1,4 +1,5 @@
 import os
+import textwrap
 import sys
 
 import esm_environment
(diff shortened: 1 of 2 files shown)


dbarbi (approved):

dbarbi:
will merge this right now

PR #75, opened 2021-02-11
Reusable_filetypes in config files

mandresm:
Allows to define the `reusable_filetypes` variable inside the general section of a simulation. By default the only reusable files are `bin` and `src`. With this change it is possible to also reuse, for example, `input`. This means that the simulation will copy the input once into the experiment folder, and will use copies or links of this input in the experiment folder (`<exp_path>/input`) for the following legs/runs.

Additionally, another fix is included: Reused subfolders can be now copied correctly into the work directory (before it was making a mess)

### DIFF
--- a/esm_runscripts/filelists.py
+++ b/esm_runscripts/filelists.py
@@ -154,7 +154,7 @@ def reuse_sources(config):
                     config[model][filetype + "_sources"][categ] = (
                         config[model]["experiment_" + filetype + "_dir"]
                         + "/"
-                        + config[model][filetype + "_targets"][categ].split("/")[-1]
+                        + config[model][filetype + "_targets"][categ]
                     )
     return config
 
--- a/esm_runscripts/prepare.py
+++ b/esm_runscripts/prepare.py
@@ -442,7 +442,7 @@ def _add_all_folders(config):
         "config",
         "restart_in",
     ]
-    config["general"]["reusable_filetypes"] = ["bin", "src"]
+    config["general"]["reusable_filetypes"] = config["general"].get("reusable_filetypes", ["bin", "src"])
 
     config["general"]["thisrun_dir"] = (
         config["general"]["experiment_dir"]


esm_runscripts/filelists.py
```diff
@@ -154,7 +154,7 @@ def reuse_sources(config):
                     config[model][filetype + "_sources"][categ] = (
                         config[model]["experiment_" + filetype + "_dir"]
                         + "/"
-                        + config[model][filetype + "_targets"][categ].split("/")[-1]
+                        + config[model][filetype + "_targets"][categ]
```
pgierz, line 157: I don't understand this change.

esm_runscripts/prepare.py
```diff
@@ -442,7 +442,7 @@ def _add_all_folders(config):
         "config",
         "restart_in",
     ]
-    config["general"]["reusable_filetypes"] = ["bin", "src"]
+    config["general"]["reusable_filetypes"] = config["general"].get("reusable_filetypes", ["bin", "src"])
```
pgierz, line 445: Looks good. That means, it is taken either from the user runscript, or from the setup yaml, right?

esm_runscripts/filelists.py
mandresm: The old line allowed only the reuse of resources that are files, but not folders.

The change in this line now allows the reuse also of subfolders.

Let's say for example that you want to include `input` in the reuse sources and your input has subfolders:
```
<exp_path>/input/model/subfolder/many_files
```
If you don't remove the `.split("/")[-1]` you'll end up with `many_files` directly on work, where you might want to have those files in `work/subfolder` instead. Also, this was resulting in an even more complicated and wrong behaviour when combined with the following subfolder hotfixes:
- #69 
- #72 

There, the subfolders would be copied but the files contained in each should folder would be those of the first subfolder only, or something crazy like that.

esm_runscripts/prepare.py
mandresm: Correct. For example in awicm3.yaml:
https://github.com/esm-tools/esm_tools/blob/feature/awicm3/configs/setups/awicm3/awicm3.yaml#L40

esm_runscripts/filelists.py
pgierz: OK, cool. thanks for explaining.

pgierz (approved):

PR #80, opened 2021-02-16
Feature/esm tools dir class

mandresm:
This merge uses the new `esm_rcfile.EsmToolsDir` class to cleanup the code from the lines that evaluate whether the run is open or in a virtual environment, and assigns the correct paths to `esm_tools` package directories.

Tested for AWICM-3 and AWICM-1, both for open run and virtual environment.

See testing comments in esm-tools/esm_parser#25


dbarbi:
will be merged now

PR #84, opened 2021-02-26
feature/model specific reusable filetypes

mandresm:
added the possibility to control the reusable_filetypes both from each component and from the general section

### DIFF
--- a/esm_runscripts/filelists.py
+++ b/esm_runscripts/filelists.py
@@ -155,7 +155,7 @@ def reuse_sources(config):
     all_reusable_filetypes = []
     for model in config["general"]["valid_model_names"] + ["general"]:
          all_reusable_filetypes = list(
-            set(all_reusable_filetypes) | set(config[model].get("reusable_filetypes", []))
+            set(all_reusable_filetypes + config[model].get("reusable_filetypes", []))
         )
     # Loop through all the reusable file types
     for filetype in all_reusable_filetypes:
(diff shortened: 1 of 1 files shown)


esm_runscripts/filelists.py
```diff
@@ -147,9 +147,28 @@ def complete_sources(config):
...
-    for filetype in config["general"]["reusable_filetypes"]:
+
+    # MA: the changes below are to be able to specify model specific reusable_filetypes
+    # without changing the looping order (a model loop nested inside a file-type loop)
+
+    # Put together all the possible reusable file types
+    all_reusable_filetypes = []
+    for model in config["general"]["valid_model_names"] + ["general"]:
+         all_reusable_filetypes = list(
+            set(all_reusable_filetypes) | set(config[model].get("reusable_filetypes", []))
```
denizural: I think you are trying to eliminate the duplicates by using the `set` datatype here. Maybe this part can be made shorter like:
```python
all_reusable_filetypes = list ( set( all_reusable_filetypes.extend(config[model].get("reusable_filetypes", [])) ) ) 
```
or even shorter with the overloaded `+` operator:
```python
all_reusable_filetypes = list ( set( all_reusable_filetypes + config[model].get("reusable_filetypes", []) ) ) 
```
Here the list extend / concatenation, merges two lists together and eliminates casting both lists to a set.

denizural (approved):

esm_runscripts/filelists.py
mandresm: I agree, these lines are better programmed in the way you explained. I'll change that to the last of the solutions you provided. Thanks :)

PR #86, opened 2021-03-15
heterogeneous paralellization (give each executable its own omp_num_threads)

JanStreffing:
This is the first time I work on the backend. Please take extra care that I did not violate any basic rules.

### DIFF
--- a/esm_runscripts/slurm.py
+++ b/esm_runscripts/slurm.py
@@ -87,7 +87,7 @@ class Slurm:
         else:
             return start_proc, start_core, end_proc, end_core
 
-        if "taskset" in config["general"]:
+        if "taskset" in config["general"] == "true":
             command = "./" + config[model]["execution_command_script"]
             scriptname="script_"+model+".ksh"
             with open(self.folder+scriptname, "w") as f:
--- a/esm_runscripts/slurm.py
+++ b/esm_runscripts/slurm.py
@@ -16,6 +16,8 @@ class Slurm:
     ----------
     filename : str
         The filename for srun commands, defaults to ``hostfile_srun``
+    hostlist : str
+        The hostlist for srun commands, defaults to ``hostlist``
     path : str
         Full path to this file, defaults to ``thisrun_scripts_dir / filename``
 
@@ -28,6 +30,7 @@ class Slurm:
     def __init__(self, config):
         folder = config["general"]["thisrun_scripts_dir"]
         self.filename = "hostfile_srun"
+        self.hostlist = "hostlist"
         self.path = folder + "/" + self.filename
 
     @staticmethod
@@ -69,7 +72,10 @@ class Slurm:
     @staticmethod
     def mini_calc_reqs(config, model, hostfile, start_proc, end_proc):
         if "nproc" in config[model]:
-            end_proc = start_proc + int(config[model]["nproc"]) - 1
+            if "OMP_NUM_PROC" in config[model]:
+                end_proc = start_proc + int(config[model]["nproc"])*int(config[model]["OMP_NUM_PROC"]) - 1
+            else:
+                end_proc = start_proc + int(config[model]["nproc"]) - 1
         elif "nproca" in config[model] and "nprocb" in config[model]:
             end_proc = start_proc + int(config[model]["nproca"])*int(config[model]["nprocb"]) - 1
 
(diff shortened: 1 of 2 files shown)


esm_runscripts/slurm.py
```diff
@@ -80,16 +83,40 @@ def mini_calc_reqs(config, model, hostfile, start_proc, end_proc):
                     end_proc += config[model]["nprocar"] * config[model]["nprocbr"]
 
         else:
-            return start_proc, end_proc
+            return start_proc, start_core, end_proc, end_core
+
+        if config["general"].get("taskset", False): 
+            if "taskset" in config["general"]:
```
mandresm: I think this `if` is not necessary here

esm_runscripts/slurm.py
```diff
@@ -29,6 +29,7 @@ def __init__(self, config):
         folder = config["general"]["thisrun_scripts_dir"]
         self.filename = "hostfile_srun"
         self.path = folder + "/" + self.filename
+        self.folder = folder + "/../work/"
```
mandresm: Why is the work folder needed here? Is it not possible to use the same logic as with the `hostfile_srun`?

esm_runscripts/slurm.py
```diff
@@ -100,10 +127,11 @@ def calc_requirements(self, config):
...
         start_proc = 0
-        end_proc = 0
-        with open(self.path, "w") as hostfile:
-            for model in config["general"]["valid_model_names"]:
-                start_proc, end_proc = self.mini_calc_reqs(config, model, hostfile, start_proc, end_proc)
+        start_core = 0
+        end_proc = 0 
+        end_core = 0
+        for model in config["general"]["valid_model_names"]:
+            start_proc, start_core, end_proc, end_core = self.mini_calc_reqs(self,config, model, start_proc, start_core, end_proc, end_core)
```
mandresm: Is it `self` required as an input?

esm_runscripts/slurm.py
```diff
@@ -67,9 +68,11 @@ def calc_requirements_multi_srun(self, config):
 
 
     @staticmethod
-    def mini_calc_reqs(config, model, hostfile, start_proc, end_proc):
+    def mini_calc_reqs(self,config, model, start_proc, start_core, end_proc, end_core):
```
pgierz: `self` shouldn't be an argument to static methods.

JanStreffing: Not sure I understand what's wrong here. Do I extract path from self.path and give that as an argument?

esm_runscripts/slurm.py
```diff
@@ -130,8 +129,9 @@ def calc_requirements(self, config):
         start_core = 0
         end_proc = 0 
         end_core = 0
+        path=self.path
```
pgierz: you can combine this into the line below.

```
return_things = self.mini_calc_reqs(self.path, config, ....)
```

JanStreffing: Probably a good sign to call it quits for the day xD

esm_runscripts/slurm.py
```diff
@@ -67,9 +67,11 @@ def calc_requirements_multi_srun(self, config):
 
 
     @staticmethod
-    def mini_calc_reqs(config, model, hostfile, start_proc, end_proc):
+    def mini_calc_reqs(path, config, model, start_proc, start_core, end_proc, end_core):
```
mandresm: This will break `start_proc, end_proc = self.mini_calc_reqs(config, model, hostfile, start_proc, end_proc)` in the method `calc_requirements_multi_srun`.

I also don't like the idea of not passing a file object (e.g. `hostfile`) to the `mini_calc_reqs` method and using `with open('<file>', 'a') afile:` inside `mini_cal_reqs` instead. Note that if you pass the file object as it was done before (see the next comment) you can use `'w'` when you open the file outside of the method which rewrites whatever file already existed. If you have it inside this method and you use `'a'` for `open` the file it won't be possible to use multiple times `esm_runscripts` without cleaning the experiment directory, as this will be appending lines to an already existing file from the previous `esm_runscript` call, I would expect this to be a source of errors.

esm_runscripts/slurm.py
```diff
@@ -80,16 +82,40 @@ def mini_calc_reqs(config, model, hostfile, start_proc, end_proc):
...
-        hostfile.write(str(start_proc) + "-" + str(end_proc) + "  " + command + "\n")
-        start_proc = end_proc + 1
-        return start_proc, end_proc
+            return start_proc, start_core, end_proc, end_core
+
+        with open(path, "a") as hostfile:
+            hostfile.write(str(start_proc) + "-" + str(end_proc) + "  " + command + "\n")
+            start_proc = end_proc + 1
+            start_core = end_core + 1
+        return start_proc, start_core, end_proc, end_core
```
mandresm: As mentioned in the previous comment, I would rather see this done outside of this function.

mandresm (approved):

PR #93, opened 2021-03-26
Hotfix/wildcards renaming to target

mandresm:
A fix to allow renaming during copying into targets when using wildcards.

For example, in awicm3 we have the following restart files:
```
<exp_path>/restart/oifs/DATE/srf0006.*
```
and we want them to end in the next run work directory as `srf0003.*` such as:
```
<exp_path>/run_DATE/work/srf0003.*
```

This fix allows this type of operation as long as the source key in the yaml file has a consistent number of `*` with the target key.

### DIFF
--- a/esm_runscripts/filelists.py
+++ b/esm_runscripts/filelists.py
@@ -269,7 +269,7 @@ def target_subfolders(config):
                                 + "/"
                                 + source_filename.split("/")[-1]
                             )
-                        elif "*" in filename:
+                        else:
                             # Get the general description of the filename
                             gen_descr = descr.split("_glob_")[0]
                             # Load the wild cards from source and target and split them
@@ -293,15 +293,11 @@ def target_subfolders(config):
                             # Loop through the pieces of the wild cards to create
                             # the correct target name by substituting in the source
                             # name
-                            target_name = source_filename.split("/")[-1]
+                            target_name = source_filename
                             for wcs, wct in zip(wild_card_source, wild_card_target):
                                 target_name = target_name.replace(wcs, wct)
                             # Return the correct target name
                             config[model][filetype + "_targets"][descr] = target_name
-                        else:
-                            config[model][filetype + "_targets"][
-                                descr
-                            ] = source_filename.split("/")[-1]
                     elif filename.endswith("/"):
                         source_filename = os.path.basename(
                             config[model][filetype + "_sources"][descr]
--- a/esm_runscripts/filelists.py
+++ b/esm_runscripts/filelists.py
@@ -207,6 +207,12 @@ def globbing(config):
                 ):  # * only in targets if denotes subfolder
                     if "*" in filename:
                         del config[model][filetype + "_sources"][descr]
+                        # Save the wildcard string for later use when copying to target
+                        wild_card = config[model].get(
+                            f"{filetype}_sources_wild_card", {}
+                        )
+                        wild_card[descr] = filename.split("/")[-1]
+                        config[model][filetype + "_sources_wild_card"] = wild_card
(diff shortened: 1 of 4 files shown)


esm_runscripts/filelists.py
```diff
@@ -258,13 +264,40 @@ def target_subfolders(config):
...
                         elif "/" in filename:
-                            config[model][filetype + "_targets"][
-                                descr
-                            ] = "/".join(filename.split("/")[:-1]) + "/" + source_filename.split("/")[-1]
+                            config[model][filetype + "_targets"][descr] = (
+                                "/".join(filename.split("/")[:-1])
+                                + "/"
+                                + source_filename.split("/")[-1]
+                            )
                         else:
```
denizural, line 272: Overall fix looks very nice 👏 👏 
However, I see one issue here:
This is a fix for the `<exp_path>/restart/oifs/DATE/srf0006.*` cases, where the file name ends with the glob wildcard. `else` branch would catch this but what I would do is something like this
```python
elif filename.endswith('.*'):    # NOTE-1
    # ...
    # rest of the hotfix
    # ...
else:
    # ...
    # these would get executed if nothing else is matched (like a case default situation). 
    # Eg, it could trap some typos with the wildcards
    # Maybe a warning message or a placeholder for a further bugfix
```
NOTE-1: to be on the safe side a regex pattern of `re.search("\.\*[ \t]*$", filename)` could have been used but it could also be an overkill

esm_runscripts/filelists.py
```diff
@@ -258,13 +264,40 @@ def target_subfolders(config):
...
+                            # Load the wild cards from source and target and split them
+                            # at the *
+                            wild_card_source_all = config[model][
+                                f"{filetype}_sources_wild_card"
+                            ]
+                            wild_card_source = wild_card_source_all[gen_descr].split(
+                                "*"
+                            )
+                            wild_card_target = filename.split("*")
+                            if len(wild_card_target) != len(wild_card_source):
```
seb-wahl, line 285: Why do they have to have the same length. Shouldn't we also allow
`<exp_path>/restart/oifs/DATE/srf0006.*`
and we want them to end in the next run work directory as `crazy_new_name_*` such as: `<exp_path>/run_DATE/work/crazy_new_name_*`
Or is there another reason for the limitation I might have overseen?

seb-wahl (changes requested):

esm_runscripts/filelists.py
mandresm: Not the same length, but the same number of pieces if you slice them with "*".
For example:
- `asdasd*asdasd*mmmm` copied to `a*a*m` is allowed
- `asdasd*asdasd*mmmm` copied tp `a*a` is not allowed

esm_runscripts/filelists.py
mandresm: I think what you are asking for is already solved here:
https://github.com/esm-tools/esm_runscripts/blob/hotfix/wildcards_renaming_to_target/esm_runscripts/filelists.py#L275-L298

The idea is that the wild cards are split in pieces using the "*" so for example `asdasd*mmm` would be split in `["asdasd", "mmm"]` and then substitution of this pieces occur on the target.

esm_runscripts/filelists.py
mandresm: Note this line:  `wild_card_target = filename.split("*")`. `wild_card_target` is not a string but a list of strings resulting from slicing the wild card with the "*". I hope this is a better clarification to your question.

mandresm: I talked to @seb-wahl about it and he is find with this now.

PR #94, opened 2021-03-29
Feature/prev run

mandresm:
This pull request adds the `RunPrevInfo` class to `sim_objects.py`, a dictionary subclass to access information from the previous run. The object is created in the ``SimulationSetup`` class in ``self.config["prev_run"]``. The idea behind this class is that variables from the previous simulation can be called from the yaml files with a very similar syntax as one would do for the current run.

The syntax is as follows:
```
<your_var>: ${prev_run.<path>.<to>.<var>}
```
For example, let's assume we want to access the `time_step` from the previous run of a `FESOM` simulation and store it in a variable called `prev_time_step`:
```
prev_time_step: ${prev_run.fesom.time_step}
```
Note: Only the single previous simulation loaded

This is important for AWICM3, for copying and renaming restart files when OIFS is cheated to think it is always running the second leg, otherwise it meets a memory problem in log runs (a well known OIFS issue). The problem of cheating OIFS in this manner is that previous leg restart files names depend on some information existing in the previous run but not in the current run, so the previous config needs to be accessed somehow.

### DIFF
--- a/esm_runscripts/compute.py
+++ b/esm_runscripts/compute.py
@@ -1,6 +1,7 @@
 import os
 import shutil
 import subprocess
+import copy
 
 import esm_rcfile
 import six
@@ -324,7 +325,10 @@ def _write_finalized_config(config):
         + "_finished_config.yaml",
         "w",
     ) as config_file:
-        out = yaml.dump(config)
+        # Avoid saving ``prev_run`` information in the config file
+        config_final = copy.deepcopy(config)
+        del config_final["prev_run"]
+        out = yaml.dump(config_final)
         out = strip_python_tags(out)
         config_file.write(out)
     return config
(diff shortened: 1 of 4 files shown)


JanStreffing:
That's a whopper. I can't really review much other than whether it functions as expected for AWICM3. I will do just that. 
I plan to run tests where I 


- [ ] change the length of  the leg mid simulation.
- [ ] kill a run to simulation a hardware failure and test whether it still does everything correctly again next run.

mandresm:
@JanStreffing , you'll need the following branches for that:
- esm_parser: `feature/prev_run`
- esm_tools: `eternal_restarts_for_oifs_cherrypick_no_develop`

JanStreffing:
After lunch break :-)

denizural (approved):
Looks very nice. Great job 👏

PR #97, opened 2021-04-08
pbs.py file added, small changes and cleanup in batch_system.py

mandresm:
Support for PBS.

**Other packages**
- esm-tools/esm_tools#321

**TODO**

- [x] Check this does not break things in slurm computers
- [x] Get AWICM3 running in ALEPH
- [x] Document `pbs.py`
- [x] Finish `Pbs.get_job_state` method
- [x] Move OpenMPI-slurm implementation from `batch_system.py` into `slurm.py`.
- [x] Test OpenMPI moved arround

### DIFF
--- a/esm_runscripts/pbs.py
+++ b/esm_runscripts/pbs.py
@@ -1,0 +1,152 @@
+"""
+Contains functions for dealing with PBS-based batch systems
+"""
+import os
+import subprocess
+import sys
+import esm_parser
+
+class Pbs:
+    """
+    Deals with PBS, allowing you to check if a job is submitted, get the
+    current job ID, generate the MOAB commands (e.g. ``aprun``), get the current job
+    state, and check if a job is still running.
+
+
+    Attributes
+    ----------
+    filename : str
+        The filename for srun commands, defaults to ``hostfile_srun``
+    path : str
+        Full path to this file, defaults to ``thisrun_scripts_dir / filename``
+
+    Parameters
+    ----------
+    config : dict
+        The run configuration, needed to determine where the script directory
+        for this particular run is.
+    """
+    def __init__(self, config):
+        # No hostfile for PBS
+        self.filename = ""
+        self.path = ""
+
+    @staticmethod
+    def check_if_submitted():
+        """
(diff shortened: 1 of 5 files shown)


esm_runscripts/pbs.py
```diff
@@ -0,0 +1,152 @@
...
+    @staticmethod
+    def get_jobid():
+        """
+        Gets the current PBS JOB ID
+
+        Returns
+        -------
+        str or None
+        """
+        return os.environ.get("PBS_JOBID")
```
denizural, line 54: Do we need a default return value here? It would be None anyway,s if it is not submitted, I guess.

esm_runscripts/pbs.py
```diff
@@ -0,0 +1,152 @@
...
+    def calc_requirements(self, config):
+        """
+        Calculates requirements and writes them to ``self.path``.
+        """
+        # PBS does not support yet multi_apruns
+        #if config['general'].get('multi_apruns'):
+        #    self.calc_requirements_multi_aprun(config)
+        #    return
+        component_lines = []
+        sep = config["computer"].get("launcher_comp_sep", "\\\n    ") + " "
```
denizural, line 176: I think these lines require little bit explanation for the others. `sep` looks like a separator, right?

mandresm: Correct, bare in mind this is still a Draft Pull Request, I will comment everything later on, when I do the cleanup

seb-wahl:
I couldn't find anything that made me worry. 
It reminds me that my quick integration of support for `mpirun` is not very clean as I misuse the `hostile_srun`. Once you PBS stuff is merged, and if we really use mpirun on more machines, I'll see if we can get a cleaner integration of `mpirun`.

seb-wahl (approved):
see my comment in the conversation

JanStreffing:
I can not comment on style. I've seen that the latest version works with pbs on aleph. I assume it also works with slurm still, since all the slurm things have gone into the slurm.py now?

JanStreffing (approved):

PR #104, opened 2021-04-29
minifeature/pre run commands

mandresm:
added the variable `pre_run_commands` to specify commands on the .sad file to be written before the run calls. It can be defined inside any component and inside any yaml file as either a string or a list of strings

### DIFF
--- a/esm_runscripts/batch_system.py
+++ b/esm_runscripts/batch_system.py
@@ -5,6 +5,7 @@ import sys
 import esm_environment
 import six
 
+from esm_parser import user_error
 from . import helpers
 from .slurm import Slurm
 from .pbs import Pbs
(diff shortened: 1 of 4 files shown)


esm_runscripts/batch_system.py
```diff
@@ -192,6 +193,32 @@ def get_extra(config):
             extras.append("source "+config["general"]["experiment_dir"]+"/.venv_esmtools/bin/activate")
         if config["general"].get("funny_comment", True):
             extras.append("# 3...2...1...Liftoff!")
+        # Search for ``pre_run_commands``s in the components
+        for component in config.keys():
+            pre_run_commands = config[component].get("pre_run_commands")
```
seb-wahl, line 198: does this mean that pre_run_commands cannot go into the `general:` section?

seb-wahl (approved):

esm_runscripts/batch_system.py
mandresm: It can go in the `general` section. I think the `component` name is misleading, config contains `general`, `computer`, `foci` ... so we are looping through `general` too.

denizural (approved):
Looks pretty solid! 👏

PR #108, opened 2021-05-03
hotfix/crash stop

mandresm:
fix the esm_runscripts continuing the simulation even if a run crashed

### DIFF
--- a/esm_runscripts/tidy.py
+++ b/esm_runscripts/tidy.py
@@ -120,7 +120,7 @@ def assemble_error_list(config):
     gconfig = config["general"]
     known_methods = ["warn", "kill"]
     stdout = (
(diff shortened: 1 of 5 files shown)


esm_runscripts/tidy.py
```diff
@@ -120,7 +120,7 @@ def assemble_error_list(config):
     gconfig = config["general"]
     known_methods = ["warn", "kill"]
     stdout = (
-        gconfig["thisrun_scripts_dir"]
+        gconfig["experiment_scripts_dir"]
```
mandresm, line 123: This are the only real changes needed for fixing the `error stop` feature. The other changes are already merged into release and I don't know why they are showing up here.

PR #111, opened 2021-05-03
added list support in batch flags

denizural:
As @seb-wahl asked on April 29th on small_technical_stuff channel, lists are not supported in the batch commands. So that it is not possible to write

```yaml
additional_flags:
- '--foo'
- '--bar'
```
and get 
```bash
#SBATCH --foo
#SBATCH --bar
```
in the `sad` file. This PR is a feature that allows lists to be used. The changes are only in `get_batch_header`  function but the same logic can be applied to other functions as well.

### DIFF
--- a/esm_runscripts/batch_system.py
+++ b/esm_runscripts/batch_system.py
@@ -93,17 +93,37 @@ class batch_system:
         ]
         if config["general"]["jobtype"] in ["compute", "tidy_and_resume"]:
             conditional_flags.append("exclusive_flag")
+
         for flag in conditional_flags:
-            if flag in this_batch_system and not this_batch_system[flag].strip() == "":
-                all_flags.append(flag)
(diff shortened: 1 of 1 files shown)


esm_runscripts/batch_system.py
```diff
@@ -93,17 +93,37 @@ def get_batch_header(config):
         ]
         if config["general"]["jobtype"] in ["compute", "tidy_and_resume"]:
             conditional_flags.append("exclusive_flag")
+
         for flag in conditional_flags:
```
mandresm, line 99: I think this whole loop takes care of getting rid of the empty flags. If that is correct I would add a comment here:
`# Avoid using empty conditional flags (strings and list of strings supported)`.

mandresm (approved):
Great improvement, I see no issues.

PR #120, opened 2021-05-06
symlinks: may the force be with you

denizural:
As Sebastian pointed out on Slack, we need force in symbolic linking (aka ln -sf). Eg. 
```bash
ln -s target link
ln -s target link  # this does not work: 'ln: creating symbolic link `link': File exists'
ln -sf target link  # but this does
```
Unfortunately Python `os.symlink` does not have force/overwrite option. While writing my own I found a upvoted solution on SO. I tested that on a separate code and seems to work.

I placed this function into `helpers.py`. esm_runscripts have many more `os.symlink` calls but I did not change them yet. I only changed the part that caused trouble to @seb-wahl.

### DIFF
--- a/esm_runscripts/helpers.py
+++ b/esm_runscripts/helpers.py
@@ -7,6 +7,7 @@ import esm_rcfile
 
 import esm_plugin_manager
 import esm_tools
+import esm_parser
 
 
 def symlink(target, link_name, overwrite=False):
@@ -20,7 +21,12 @@ def symlink(target, link_name, overwrite=False):
     '''
 
     if not overwrite:
-        os.symlink(target, link_name)
+        try:
+            os.symlink(target, link_name)
+        except FileExistsError:
+            error_type = "File Exists"
+            error_text = f"{link_name} already exists. Use overwrite=True"
+            esm_parser.user_error(error_type, error_text, exit_code=1)
         return
 
     # os.replace() may fail if files are on different filesystems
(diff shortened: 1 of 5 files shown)


esm_runscripts/helpers.py
```diff
@@ -1,12 +1,56 @@
...
+    # Replace link_name with temp_link_name
+    try:
+        # Pre-empt os.replace on a directory with a nicer message
+        if not os.path.islink(link_name) and os.path.isdir(link_name):
+            raise IsADirectoryError(f"Cannot symlink over existing directory: '{link_name}'")
+        os.replace(temp_link_name, link_name)
+    except:
+        if os.path.islink(temp_link_name):
+            os.remove(temp_link_name)
+        raise
```
seb-wahl: I personally don't like the raise without anything else. It will print a horrible python error to the user who will then start crying and approach one of us :-) No seriously, we should either use
1. @mandresm 's `user_error` function or
2. print a clean error message, and do `sys.exit(1)`

seb-wahl (changes requested):

esm_runscripts/helpers.py
denizural: Totally, agree. This is also what I thougt. Bare exceptions are not good. Let me fix that.

mandresm: Nice fix @denizural !

I would use `esm_parser.user_error` as @seb-wahl suggested, it already does the `sys.exit(1)` for you and it gives a consistent look to errors. It takes a title and a descriptive text as input and generates the following message:

```
ERROR: <the title>
-------------------

Your descriptive text
```

seb-wahl (approved):

mandresm (approved):

PR #122, opened 2021-05-10
Feature/handle sigint

denizural:
As @dbarbi mentioned today, pressing `CTRL-C` may result in an ugly list of traceback which is not something that we see in general Linux commands. I wrote a event listener and handler to solve that. Now, when you hit `CTRL-C` while `esm_runscripts` is running this will result in 
```bash
^C
ctrl c is pressed. Terminating
```

In the first commit, I actually placed some verbosity messages but then deleted them. They printed out the last element in the stack trace. If you don't want anything at all, then simply erase the `sys.stderr.write()` call.

If you hit `CTRL-C` in a nanosecond of course it will not be executed. I can also make a workaround but this requires some changes to the `setup.py`. There we need to provide our own binary executable instead of the default one that is provided by Python. I mean, we need to use `scripts` explicitly instead of the `console_scripts` `entry_points`. Then, I can explicitly have something like
```Python
#!/sw/rhel6-x64/conda/anaconda3-bleeding_edge/bin/python
# 1st line to execute in esm_runscripts from command line
import event_handlers
event_handlers.signal_listener()

# rest of it is the same as before as automatically generated by Python package builder

__requires__ = 'esm-runscripts'
import re
import sys
from pkg_resources import load_entry_point

if __name__ == '__main__':
    sys.argv[0] = re.sub(r'(-script\.pyw?|\.exe)?$', '', sys.argv[0])
    sys.exit(
    )
```

It is also possible to extend this implementation for the other signals such as `CTRL-z` and so on. I can also use the samething in our other binaries such as `esm_master` and `esm_versions`. But for that, I think it is better to move these commands to a more common path for avoiding duplication.

### DIFF
--- a/esm_runscripts/event_handlers.py
+++ b/esm_runscripts/event_handlers.py
@@ -28,11 +28,6 @@ def handle_sigint(signum, stack_frame):
     print()  # print a new line after ^C character (CTRL-C)
     sys.stderr.write('ctrl c is pressed. Terminating\n')
 
-    # these can be moved into a `if verbose` block
-    sys.stderr.write(f'- file:     {os.path.realpath(stack_frame.f_code.co_filename)}\n')
-    sys.stderr.write(f'- line:     {stack_frame.f_code.co_firstlineno}\n')
-    sys.stderr.write(f'- function: {stack_frame.f_code.co_name}\n')
-    
     # exit with 2 according to https://en.wikipedia.org/wiki/Signal_(IPC) 
     sys.exit(signal.SIGINT.value)  
             
--- a/esm_runscripts/event_handlers.py
+++ b/esm_runscripts/event_handlers.py
@@ -23,7 +23,9 @@ def handle_sigint(signum, stack_frame):
     stack_frame : Python frame object
         current stack frame
     
-    More info: https://docs.python.org/3/library/signal.html
+    Notes
+    -----
+        More info: https://docs.python.org/3/library/signal.html
     """
     print()  # print a new line after ^C character (CTRL-C)
     sys.stderr.write('ctrl c is pressed. Terminating\n')
--- a/esm_runscripts/event_handlers.py
+++ b/esm_runscripts/event_handlers.py
@@ -1,0 +1,41 @@
+import signal
+import sys
+import os
+
+def signal_listener():
+    """Listens to the signals (eg. CTRL-C press, ...) and calls the handler
+    for that signal"""
+    
+    # 1) handler for the SIGINT (CTRL-C)
+    signal.signal(signal.SIGINT, handle_sigint)
+    
+    # other signal handlers may go here, eg. SIGHUP, SIGSTOP, ...)
+    # each signal is handled by a function handle_<signal>, eg. handle_sigint()
+
+
+def handle_sigint(signum, stack_frame):
+    """Handles the SIGINT signal (aka CTRL-C press on UNIX/LINUX)
+    
+    Parameters
+    ----------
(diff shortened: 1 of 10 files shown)


esm_runscripts/event_handlers.py
```diff
@@ -0,0 +1,36 @@
...
+    """Handles the SIGINT signal (aka CTRL-C press on UNIX/LINUX)
+    
+    Parameters
+    ----------
+    signum : int
+        signal number.
+    stack_frame : Python frame object
+        current stack frame
+    
+    More info: https://docs.python.org/3/library/signal.html
```
pgierz: Stick that under a Notes heading:
```
Notes
------
The signal library is documented here: https://docs.python.org/3/library/signal.html
```

esm_runscripts/event_handlers.py
```diff
@@ -0,0 +1,36 @@
...
+    """
+    print()  # print a new line after ^C character (CTRL-C)
+    sys.stderr.write('ctrl c is pressed. Terminating\n')
+
+    # exit with 2 according to https://en.wikipedia.org/wiki/Signal_(IPC) 
+    sys.exit(signal.SIGINT.value)  
+            
+
+if __name__ == "__main__":
```
pgierz, line 37: Not technically needed, since everything in this file is just definitions and declaration.

esm_runscripts/event_handlers.py
```diff
@@ -0,0 +1,36 @@
...
+import os
+
+def signal_listener():
+    """Listens to the signals (eg. CTRL-C press, ...) and calls the handler
+    for that signal"""
+    
+    # 1) handler for the SIGINT (CTRL-C)
+    signal.signal(signal.SIGINT, handle_sigint)
+    
+    # other signal handlers may go here, eg. SIGHUP, SIGSTOP, ...)
```
pgierz, line 12: `CTRL-Z + bg` might be nice to launch the job, keep the launch going in the background, and getting your shell back. We might need a way to suppress any sort of printing in the case so you don't get those distributing the rest of your work.

pgierz (commented):
The code changes look good, only some minor ideas.

I didn't fully understand what you mean by "we would need to re-write the `setup.py`", maybe someone can explain it to me in the meeting next time.

esm_runscripts/event_handlers.py
denizural: Here is the file call hierarchy: 

1. first you call the executable (esm_runscripts, eg. ~/.local/bin/esm_runscripts).
2. This calls `load_entry_point('esm-runscripts', 'console_scripts', 'esm_runscripts')()`
3. This calls `'console_scripts': ['esm_runscripts=esm_runscripts.cli:main', ],`
4. Enter `cli.py` that I modified.

Obviously, it takes some time until the execution reaches `cli.py`. But what happens if someone executes `esm_runscripts` can then presses `CTRL-C` within few seconds. It will print the nasty error stack since the event listener has not started yet. 

On the other hand, if I place the `event_handler` to the top of the `esm_runscripts` CLI executable (that is automatically created by the Python setuptools module then there will be a greater chance to catch that early `CTRL-C`.

According to the setuptools documentation, it will generate an automatic CLI executable if you use the `console_scripts` entry point. You can provide your own executable script if you use `scripts` instead.

Here is a nice read: https://python-packaging.readthedocs.io/en/latest/command-line-scripts.html

This is a personal preference of course. Sometimes I hit `CTRL-C` within 1-2 seconds since I realize that I forgot something

PR #123, opened 2021-05-10
supressing of MOTD

denizural:
As @dbarbi mentioned today in our weekly meeting, `esm_runscripts` printing the MOTD twice is quite annoying. To prevent that a command line argument `--no-motd` can be provided. If you provide that when running `esm_runscripts` directly then no MOTD will be printed at all. If you don't then MOTD will be printed once but for the subsequent cases it will not be printed.

I think it is a good thing to have a flag to supress MOTD since it is annoying if you run `esm_master` or `esm_runscripts` many times.

### DIFF
--- a/esm_runscripts/__init__.py
+++ b/esm_runscripts/__init__.py
@@ -2,7 +2,7 @@
 
 __author__ = """Dirk Barbi"""
 __email__ = 'dirk.barbi@awi.de'
-__version__ = "5.1.9"
+__version__ = "5.1.10"
 
 from .sim_objects import *
 from .batch_system import *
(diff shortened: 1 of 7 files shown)


mandresm (approved):
Nice elegant fix 👍

pgierz:
I think part of it popping up twice also has to do with the virtual environment bootstrapping (the job restarts itself), so make sure to check that too, but from what I can tell that might already be handled...

pgierz (commented):
I think part of it popping up twice also has to do with the virtual environment bootstrapping (the job restarts itself), so make sure to check that too, but from what I can tell that might already be handled...

pgierz:
Can I set this in my runscript yaml as well?

denizural:
> Can I set this in my runscript yaml as well?

@pgierz, currently it is not but that sounds great and I added to my todo list. I will make a quick merge now and then I can revisit this a feature

PR #124, opened 2021-05-12
hotfix/pbs prepend model stdout stderr

mandresm:
In contrast with `srun`, `aprun` (Cray's parallel launcher) does not have a flag that allows for prepending the PE task to the stdout and stderr. This makes debugging of a coupled setup such as `AWICM3` more difficult, as it might not be clear where an error message might come from.

This hotfix prepends the corresponding `[<MODEL>]` to every stdout/stderr line when using PBS.

Before, the `aprun` call in the `.sad` file for `AWICM3` would look like:
```
time aprun -n 40 -N 40 -d 1 env OMP_NUM_THREADS=1 ./fesom \
         : -n 80 -N 40 -d 1 env OMP_NUM_THREADS=1 ./oifs -v ecmwf -e awi3 \
         : -n 1 -N 1 -d 1 env OMP_NUM_THREADS=1 ./rnfma  &
```
To prepend the `[<MODEL>]` the `aprun` call now looks like:
```
time aprun -n 40 -N 40 -d 1 env OMP_NUM_THREADS=1 bash -c "./fesom 1> >(sed 's/^/[FESOM]  /' >&1) 2> >(sed 's/^/[FESOM]  /' >&2)" \
         : -n 80 -N 40 -d 1 env OMP_NUM_THREADS=1 bash -c "./oifs -v ecmwf -e awi3 1> >(sed 's/^/[OIFS]   /' >&1) 2> >(sed 's/^/[OIFS]   /' >&2)" \
         : -n 1 -N 1 -d 1 env OMP_NUM_THREADS=1 bash -c "./rnfma 1> >(sed 's/^/[RNFMAP] /' >&1) 2> >(sed 's/^/[RNFMAP] /' >&2)"  &
```

As you can see we are substituting the call of the executable (i.e. `./fesom`) by `bash -c "./fesom <filtering commands for stdout/stderr>"`.

I find this not elegant at all but it works. Does anyone know if this change could have a negative impact in performance? Any suggestions for improvement?

### DIFF
--- a/esm_runscripts/__init__.py
+++ b/esm_runscripts/__init__.py
@@ -2,7 +2,7 @@
 
 __author__ = """Dirk Barbi"""
 __email__ = 'dirk.barbi@awi.de'
-__version__ = "5.1.8"
+__version__ = "5.1.9"
 
 from .sim_objects import *
 from .batch_system import *
--- a/esm_runscripts/pbs.py
+++ b/esm_runscripts/pbs.py
@@ -170,13 +170,6 @@ class Pbs:
         #    self.calc_requirements_multi_aprun(config)
         #    return
 
-        # Calculates the left justification for the prepend of stdout and stderr
-        model_name_lengths = [
-            len(model) for model in config["general"]["valid_model_names"]
-            if "execution_command" in config[model] or "executable" in config[model]
-        ]
-        prep_just = max(model_name_lengths) + 3
-
         component_lines = []
         # Read in the separator to be used in between component calls in the job
         # launcher
@@ -193,13 +186,9 @@ class Pbs:
             if command:
                 launcher = config["computer"].get("launcher")
                 launcher_flags = self.calc_launcher_flags(config, model)
-                # stdout and stderr modifications to prepend [MODEL] to each line
-                prep_str = f"[{model}]".upper().ljust(prep_just)
-                prep_model_std_oe = (
-                    f"1> >(sed 's/^/{prep_str}/' >&1) " +
-                    f"2> >(sed 's/^/{prep_str}/' >&2)"
-                )
-                component_lines.append(f'{launcher_flags} bash -c "./{command} {prep_model_std_oe}" ')
+                # Substitute @MODEL@ with the model name
+                launcher_flags = launcher_flags.replace("@MODEL@", model.upper())
+                component_lines.append(f'{launcher_flags} ./{command} ')
 
         # Merge each component flags and commands into a single string
         components = sep.join(component_lines)
--- a/esm_runscripts/pbs.py
+++ b/esm_runscripts/pbs.py
@@ -188,7 +188,7 @@ class Pbs:
                 launcher_flags = self.calc_launcher_flags(config, model)
                 # Substitute @MODEL@ with the model name
                 launcher_flags = launcher_flags.replace("@MODEL@", model.upper())
-                component_lines.append(f'{launcher_flags} ./{command} ')
+                component_lines.append(f"{launcher_flags} ./{command} ")
 
         # Merge each component flags and commands into a single string
         components = sep.join(component_lines)
--- a/esm_runscripts/pbs.py
+++ b/esm_runscripts/pbs.py
@@ -170,6 +170,13 @@ class Pbs:
         #    self.calc_requirements_multi_aprun(config)
         #    return
 
+        # Calculates the left justification for the prepend of stdout and stderr
+        model_name_lengths = [
+            len(model) for model in config["general"]["valid_model_names"]
+            if "execution_command" in config[model] or "executable" in config[model]
+        ]
+        prep_just = max(model_name_lengths) + 3
+
         component_lines = []
         # Read in the separator to be used in between component calls in the job
         # launcher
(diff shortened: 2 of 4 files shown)


hegish:
Do you have an idea of how bad the performance impact will be if everything is piped via sed? This would disable the default buffering of output/error streams of the queueing system, right?
Why can we not use an mpirun flag like "--tag-output" for OpenMPI?

mandresm:
> Do you have an idea of how bad the performance impact will be if everything is piped via sed? This would disable the default buffering of output/error streams of the queueing system, right?
> Why can we not use an mpirun flag like "--tag-output" for OpenMPI?

I think it is not disabling the default buffering of output/error streams of the queueing system, but I have no idea about the impact on the performance...

About the mpirun flag, I agree this would be the best option, I simply don't know the syntax for it and  did not find information about it in the aprun docs. @hegish can you point me to the right documentation?

mandresm:
@hegish , looking at the `man mpi` after `module load cray-mpich` in ALEPH I found `PMI_LABEL_ERROUT` environment variable. Turning it on provides the MPI rank for each line in stdout/stderr in the same way `--label` would do for `srun`. However, it comes with a potential decrease in performance:

```
     PMI_LABEL_ERROUT
               If set, causes PMI to pre‐pend the corresponding rank to
               each line of stdout and stderr by piping it through the PMI
               daemon for processing. Labeling is intended for debugging
               purposes and may affect performance if enabled.

               This option is only supported for use with the aprun
               launcher. Slurm does not use the PMI daemon and already
               supports a native option (e.g. srun ‐‐label). Likewise,
               setting the PMI_NO_FORK option will disable labeling since
               it disables the PMI daemon.

               The PMI output labeling feature is only available when using
               programming models that use PMI such as MPICH or SHMEM.
               Labeling is supported for use with any supported compiler,
               the C and Fortran languages, although in theory it is
               compiler and language independent. Enabling CCE output
               labeling (by setting CRAY_RANK_THREAD_PREFIX) will disable
               PMI labeling since both perform similar functions.

               Under rare circumstances output may be lost. In many such
               cases PMI will warn the user. However, in the event the
               application is terminated by force, output could be lost
               without warning if the PMI daemon is killed before the
               application.

               Note:  Labeling is not supported on Service MAMU nodes.

               Example

               $ export PMI_LABEL_ERROUT=1
               $ cc ‐o rank rank.c
               $ aprun ‐n 8 ‐N 4 ./rank
               0: PMI Label Err/Out Test
               0: Application sees rank: 0
               1: Application sees rank: 1
               2: Application sees rank: 2
               3: Application sees rank: 3
               4: Application sees rank: 4
               5: Application sees rank: 5
               6: Application sees rank: 6
               7: Application sees rank: 7
               Application 3206402 resources: utime ~0s, stime ~0s, Rss ~5352, inblocks ~0, outblocks ~0
```
What do you think? should we have this available only for debugging right? Then we can get rid of the sed-pipping I implemented and use `PMI_LABEL_ERROUT` when debugging.

hegish:
> @hegish , looking at the `man mpi` after `module load cray-mpich` in ALEPH I found `PMI_LABEL_ERROUT` environment variable. Turning it on provides the MPI rank for each line in stdout/stderr in the same way `--label` would do for `srun`. However, it comes with a potential decrease in performance:

This looks much better to me. Good find. srun -l also has a performance impact.

mandresm:
Thanks for your input @hegish 

The `aprun` call now looks like this:
```
time aprun -n 40 -N 40 -d 1 env OMP_NUM_THREADS=1 env PMI_LABEL_ERROUT_FORMAT="[FESOM]%l:" ./fesom \
         : -n 80 -N 40 -d 1 env OMP_NUM_THREADS=1 env PMI_LABEL_ERROUT_FORMAT="[OIFS]%l:" ./oifs -v ecmwf -e awi3 \
         : -n 1 -N 1 -d 1 env OMP_NUM_THREADS=1 env PMI_LABEL_ERROUT_FORMAT="[RNFMAP]%l:" ./rnfma  &
```

And the log file like this:
```
[OIFS]37:  we set OASIS_debug = 0 
[OIFS]38: (oasis_init_comp) WARNING:  With LUCIA load balance analysis 
[OIFS]38:  we set OASIS_debug = 0 
[OIFS]39: (oasis_init_comp) WARNING:  With LUCIA load balance analysis 
[OIFS]39:  we set OASIS_debug = 0 
[RNFMAP]0:
[RNFMAP]0: CPLNG: Coupling field definition.
[RNFMAP]0: CPLNG: ... Number of fields is:    4   
[OIFS]0:## EC_MEMINFO 
[OIFS]0:## EC_MEMINFO
[OIFS]0:## EC_MEMINFO Detailed memory information for program ./oifs
[OIFS]0:## EC_MEMINFO Running with    80 tasks and   1 threads at time 00:40:37.795 on 13-05-21
[OIFS]0:## EC_MEMINFO The Job Name is TCO_no_qalter_6 and the Job ID is 160830.sdb
[OIFS]0:## EC_MEMINFO 
[OIFS]0:## EC_MEMINFO               | TC    | MEMORY USED(MB) |          MEMORY FREE(MB)        INCLUDING CACHED |  %USED %HUGE
[OIFS]0:## EC_MEMINFO               | Malloc| Inc Heap        | Numa node 0    | Numa node 1    |                |   
[OIFS]0:## EC_MEMINFO Node Name     | Heap  | RSS(sum)        | Small  Huge or | Small  Huge or | Total          |   
[OIFS]0:## EC_MEMINFO               | (sum) | Small    Huge   |  Only   Small  |  Only   Small  | Memfree+Cached |
[OIFS]0:## EC_MEMINFO    1 nid00135     407    1081     400       234    6240       77    3278    186502    2408      0.8   0.0 s/p:mpl_init
[OIFS]0:## EC_MEMINFO    2 nid00266     409    1069     400       423    6106       59    3236    186530    2389      0.8   0.0 s/p:mpl_init
[FESOM]0: MPI has been initialized, provided MPI thread support level: MPI_THREAD_SINGLE 0
[FESOM]0: Running on  40  PEs 
[FESOM]0: 
[FESOM]0: FESOM2 git SHA: 72f9988
[FESOM]0: ^[[32m____________________________________________________________^[[0m
[FESOM]0: ^[[7;32m --> FESOM BUILDS UP MODEL CONFIGURATION                    ^[[0m
[OIFS]0:[EC_DRHOOK:hostname:myproc:omptid:pid:unixtid] [YYYYMMDD:HHMMSS:epoch:walltime] [function@file:lineno] -- Max OpenMP threads = 1 
```

I think it is ready to merge.

hegish:
> I think it is ready to merge.

cool!

hegish (approved):
I do not understand the ESM internals, but the resulting command looks fine for me

mandresm:
> I do not understand the ESM internals, but the resulting command looks fine for me

I moved most of the logic of this fix into the `aleph.yaml` inside a an `mpich` choose block, as this might be very specific for this configuration. Most of the changes are here:

https://github.com/esm-tools/esm_tools/blob/3c855784139661ee5dcd2cac7f993f1f456eec3e/configs/machines/aleph.yaml#L124-L130

PR #129, opened 2021-05-21
Flushed tidy print statements

pgierz:
Several of the print statements in the tidy job are showing up considerably
late, as the print statements aren't directly flushed to stdout.

This PR flushes all print statements in oasis.py and filelists.py directly.

A more general note, we should consider migrating print to logging (or, maybe,
loguru), as that will automatically allow us to add timestamps, file locations,
etc to the logged output, but this would require a larger refactor.


mandresm (approved):
I agree with you. It's being also a while that I wanted to prepare a little intro for the team about logguru, but did not find the time to prepare it. Maybe something to do for release 6?

pgierz:
@mandresm, can you take care of the bump version stuff? 

And yes, we might spend a week before release 6 and halt all the new features and just do a general cleanup.

mandresm:
> @mandresm, can you take care of the bump version stuff?

Yes, I'll do that

PR #132, opened 2021-06-10
fixed verbose in resolve_symlinks

seb-wahl:
`config["general"]["verbose"]` is not available in `resolve_symlinks` as `config` is not passed to `resolve_symlinks`. I think I was the first hitting this error when using linking to excessively.


mandresm (approved):
Looks good to me :) . Let me know if I should bump the version and merge now.

PR #133, opened 2021-06-11
fix(database_actions): allows user to opt-out of using the database

pgierz:
…See Issue https://github.com/esm-tools/esm_tools/issues/355)

This allows the user to opt-out of the database actions by setting `general.use_database` to False in their runscript. Currently the default is True


mandresm (approved):

pgierz:
Will close esm-tools/esm_tools#355

mandresm:
Done in #133

PR #140, opened 2021-06-16
added ability to read export_mode from coupling directions block

seb-wahl:
could this be merged soon, it's required for our new people to run NEMO with high resolution nest (NEMO-AGRIF) within ESM-Tools

### DIFF
--- a/esm_runscripts/coupler.py
+++ b/esm_runscripts/coupler.py
@@ -255,20 +255,23 @@ class coupler_class:
                                 if not found_left or not found_right:
                                     sys.exit(0)
 
-                            direction_info = None
-                            if "coupling_directions" in full_config[self.name]:
-                                if right_grid + "->" + left_grid in full_config[self.name]["coupling_directions"]:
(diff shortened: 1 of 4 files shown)


esm_runscripts/coupler.py
```diff
@@ -255,20 +255,23 @@ def add_couplings(self, full_config):
                                 if not found_left or not found_right:
                                     sys.exit(0)
 
+                            if "export_mode" in full_config[self.name]:
+                                export_mode = full_config[self.name]["export_mode"]
+                            else:
+                                export_mode = "DEFAULT"
```
mandresm: Can't this whole block be written as `export_mode = full_config[self.name].get("export_mode", DEFAULT)`?

mandresm (changes requested):

esm_runscripts/coupler.py
seb-wahl: I'm sure it can, since it's a hot fix I better try it first :-)

seb-wahl: tested and pushed

mandresm (approved):
I'll merge it now :)

PR #142, opened 2021-06-18
Hotfix/update infinite recursion

denizural:
These commits contain multiple bugfixes:
- prevent `esm_runscripts` to enter a recursion
- solve the problems when `expid` is a number
- making file paths more readable. Not actually a bugfix but enhances readability

Update (21/6/2021):
- Bug fix is not complete and ready to merge. 
- Related changes in the `esm_parser`: https://github.com/esm-tools/esm_parser/pull/58

### DIFF
--- a/esm_runscripts/compute.py
+++ b/esm_runscripts/compute.py
@@ -2,6 +2,7 @@ import os
 import shutil
 import subprocess
 import copy
+import pathlib
 
 import esm_rcfile
 import six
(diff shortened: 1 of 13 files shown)


esm_runscripts/compute.py
```diff
@@ -532,14 +538,33 @@ def copy_tools_to_thisrun(config):
...
+        original_command_list = original_command.split()
+        new_command_list = []
+        for command in original_command_list:
+            # current command will contain the full path, so replace it with
+            # the YAML file only since we are going to execute it from the 
+            # `scriptsdir` now
+            if runscript in command:
+                # gconfig['scriptname'] or `runscript` only contains the YAML file name
+                command = runscript 
+            new_command_list.append(command)
```
mandresm, line 559: I think I am not understanding why we need an absolute path for the runscript here. Would this not be running always from the `<exp_path/scrips>` directory?

mandresm (commented):
Nice and meticulous job :)

Just a final question, is this to be treated as a bugfix or a development?

esm_runscripts/compute.py
denizural: Hi @mandresm,

### 1) Explanation:
Unfortunately there are some hard-coded stuff in `esm_runscripts`. Take this:

in `cli.py`: 
```python
command_line_config["started_from"] = os.getcwd()
```

This is later on used in `compute.py`:
```python
# Directory where the original runscript is
fromdir = os.path.realpath(gconfig["started_from"])
# Directory where the runscript is copied to
scriptsdir = os.path.realpath(gconfig["experiment_scripts_dir"])
```
and there are checks
```python
if (fromdir == scriptsdir) and not gconfig["update"]:
```
Sure, the the `esm_runscripts` will enter the experiment directory and run the script in `<exp_path/scrips>`, in the default one this does not happen. This is why I added `gconfig['runscript_abspath']`, so that I can carry the information of where the original runscript resides.

Indeed, the cool thing is in the `restart_command`, 
`restart_command = f"cd {scriptsdir}; esm_runscripts {new_command}"`
in now reduntant. `restart_command = f"esm_runscripts {scriptsdir}/{new_command}"` should also work. So, the idea is there is no need to change the directory to the where script lies, to run it. 

### 2) Bugfix or feature:
This is a good question. To me it looks more like a bugfix since `--update` is now fixed. Calling the scripts from another directory could be more like a feature though.

mandresm: Thanks for the explanation :)

mandresm (approved):

PR #143, opened 2021-06-21
solves a problem with the check of FESOM paths from the namelist

mandresm:
The method for checking if there were missing paths in the `namelist.config` from FESOM was run only if `files_missing_when_preparing_run` exists (i.e. if there were already other missing files defined through the standard "filedictionary" way). This fix makes both checks independent, so if the missing files belong only to the `namelist.config` they are still reported.

### DIFF
--- a/esm_runscripts/__init__.py
+++ b/esm_runscripts/__init__.py
@@ -2,7 +2,7 @@
 
 __author__ = """Dirk Barbi"""
 __email__ = 'dirk.barbi@awi.de'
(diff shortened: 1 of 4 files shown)


pgierz (approved):

pgierz:
Nice catch! I always have some echam files missing in my runs (which are allowed to be missing), so I never encountered this.

As an aside: Maybe we should think about some sort of "allowed missing files" list, to avoid needlessly warning the user.

PR #145, opened 2021-06-22
hotfix/account_error

mandresm:
This solves issue #esm-tools/esm_tools#334.

This time in `esm_runscripts` where it belongs.

### DIFF
--- a/esm_runscripts/__init__.py
+++ b/esm_runscripts/__init__.py
@@ -2,7 +2,7 @@
 
 __author__ = """Dirk Barbi"""
 __email__ = 'dirk.barbi@awi.de'
-__version__ = "5.1.18"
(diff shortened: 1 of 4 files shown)


denizural (approved):
This looks nice. 

One thing that I am thinking about is maybe on some systems the user may want to run esm_tools without any submission program but only by mpirun directly. In this case this will be obsolete but I think no one uses such private computing environment.

mandresm:
Thanks for the review. Please, note that `accounting` is defined inside the different machine files and defines whether you use an account to us the computing nodes or not. 

I think that, in the case of hypothetical future systems, where the user might be allow to perform `mpirun` directly, `accounting` could be set to `False` and the new lines here will never be reached. Would you agree @denizural ?

PR #148, opened 2021-06-29
Feature/improved prev run

mandresm:
I made big changes to the `prev_run` feature, I have tested intensively so I considered it finished. I'm asking to merge into develop because I think release is not its place and I will still like to see it interact under different circumstances that have not been yet tested. I will be adding this modifications manually to `prep_release`, which should be easy as I moved the whole object to its own file `prev_run.py`.


esm_runscripts/prev_run.py
```diff
@@ -0,0 +1,493 @@
...
+            if os.path.isfile(prev_run_config_file):
+                # For debugging purposes: this line should not be reached more that
+                # once per component including ``prev_run`` variables
+                self._prev_config_count += 1
+                #print(f"PREV CONFIG COUNT: {self._prev_config_count}")
+
+                # Open the file and load the previous run information
+                with open(prev_run_config_file, "r") as prev_file:
+                    prev_config = yaml.load(prev_file, Loader=yaml.FullLoader)
+                prev_config = prev_config["dictitems"]
```
denizural, line 229: I fixed this `dictitem` part and will make PR about that

denizural: Ok, it is here https://github.com/esm-tools/esm_runscripts/pull/152
Now, you can safely remove this line since there will be no more `dictitems`, in the finished config yaml file 😀

mandresm:
It will be merged to `release` to provide the AWICM3 group with an up-to-date version of `prev_run` before release 6.

PR #151, opened 2021-06-29
fix/proc and core calc

mandresm:
EDIT: Some of the questions I had I have solved by now. **Question b** is the one keeping me away from merging into `prep_release`.

This is the first fix to `prep_release` branch, branched off from `preppreprelease`.

It consist of two different modifications:

1. `end_proc` and `end_core` were not updated when used to reset `start_proc` and `start_core` and therefore the hostfile for AWICM2 would look like:
    ```
    0-575 ./echam6
    1-288 ./fesom
    ```
    
2. This one is less clear to me and I need information from @dbarbi , @pgierz and/or @khimstedt . In `release` the `nprocar`/`nprocbr` lines are inside the `if` for `nproca`/`nprocb`:
    https://github.com/esm-tools/esm_runscripts/blob/b3a115d013cb3c553d0464527d869d0b316bf386/esm_runscripts/batch_system.py#L146-L156
    This fix makes the `if` of `nprocar`/`nprocbr` to only be evaluated if `nproca` and `nprocb` are defined (as it is the case in `release`), but in `prep_release` was made to be independent (see the diffs in the file).
I decided to come back to the `release` strategy because for my current test `nprocar` and `nprocbr` would be `0`, turning my `start_proc` to `-1`.

My questions are as follow:
    a. Should `nprocar`/`nprocbr` be independent of `nproca`/`nprocb` at the time of calculating the number of tasks/cores or should we on the contrary only check for `nprocar` if we already have an `nproca` defined?
    b. When we check the `nprocar`/`nprocbr` values, should we also check for 0 values as in the following line, or checking only for `remove_from_namelist` is okay?
    https://github.com/esm-tools/esm_runscripts/blob/326448e7e340ba8c375fdbf29cee1440b0777bf4/esm_runscripts/batch_system.py#L203

### DIFF
--- a/esm_runscripts/batch_system.py
+++ b/esm_runscripts/batch_system.py
@@ -196,28 +196,31 @@ class batch_system:
         if cluster in reserved_jobtypes:
             for model in config["general"]["valid_model_names"]:
                 omp_num_threads = int(config[model].get("omp_num_threads", 1))
-                if "nproca" in config[model] and "nprocb" in config[model]:
-                    config[model]["tasks"] = config[model]["nproca"] * config[model]["nprocb"]
+                # KH 30.04.20: nprocrad is replaced by more flexible
+                # partitioning using nprocar and nprocbr
+                no_nprocr = ["remove_from_namelist", 0]
+                if (
+                    config[model].get("nprocar", no_nprocr[0]) not in no_nprocr
+                    and
+                    config[model].get("nprocbr", no_nprocr[0]) not in no_nprocr
+                ):
+                    config[model]["tasks"] = \
+                        config[model]["nprocar"] * config[model]["nprocbr"]
+
+                elif "nproca" in config[model] and "nprocb" in config[model]:
+                    config[model]["tasks"] = \
+                        config[model]["nproca"] * config[model]["nprocb"]
                     #end_proc = start_proc + int(config[model]["nproca"])*int(config[model]["nprocb"]) - 1
-                    if (
-                        config[model].get("nprocar", "remove_from_namelist") not in ["remove_from_namelist", 0]
-                        and
-                        config[model].get("nprocbr", "remove_from_namelist") not in ["remove_from_namelist", 0]
-                        ):
-                            config[model]["tasks"] = config[model]["nprocar"] * config[model]["nprocbr"]
-                
+
                 elif "nproc" in config[model]:
                     print(f"nproc: {config[model]['nproc']}")
                     config[model]["tasks"] = config[model]["nproc"]
 
                     #cores_per_node = config['computer']['cores_per_node']
-                    # If heterogeneous MPI-OMP
-                # KH 30.04.20: nprocrad is replaced by more flexible
-                # partitioning using nprocar and nprocbr
                 else:
                         continue
 
-                    
+
                 nproc = config[model]["tasks"]
                 if cluster == "compute":
                     cores_per_node = config['computer']['partitions']['compute']['cores_per_node']
--- a/esm_runscripts/batch_system.py
+++ b/esm_runscripts/batch_system.py
@@ -196,21 +196,21 @@ class batch_system:
         if cluster in reserved_jobtypes:
             for model in config["general"]["valid_model_names"]:
                 omp_num_threads = int(config[model].get("omp_num_threads", 1))
-                # KH 30.04.20: nprocrad is replaced by more flexible
-                # partitioning using nprocar and nprocbr
-                no_nprocr = ["remove_from_namelist", 0]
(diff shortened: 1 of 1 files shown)


mandresm:
> a. Should `nprocar`/`nprocbr` be independent of `nproca`/`nprocb` at the time of calculating the number of tasks/cores or should we on the contrary only check for `nprocar` if we already have an `nproca` defined?

I changed my mind, I think they are independent and what @dbarbi did is probably the best:
1. Check if `nprocar`/`nprocbr` have a valid value and define the number of tasks
2. If not, try to find `nproca`/`nprocb` and define the number of tasks
3. If not, try to find `nproc` and define the number of tasks.

The problem I was having can be solved by checking the values of `nprocar`/`nprocbr` to don't be either `"remove_from_namelist"` or `0`.

**Question b** is still open.

khimstedt:
Hi Miguel,

 

I can provide some background information on nprocar and nprocbr that may help clarify the question.

 

1)

nprocar and nprocbr are only relevant when using the Concurrent Radiation version of ECHAM. It only makes sense to define/use them if there is also an nproca and nprocb.

 

Somewhat simplified:

a)

Standard ECHAM (or case 2) c), see below):

nproca and nprocb describe how many cores are used for ECHAM

TotalCores = nproca * nprocb

 

b)

Concurrent Radiation version of ECHAM (case 2) a) and 2) b), see below):

nproca and nprocb describe how many cores are used for the non-radiation part of ECHAM. nprocar and nprocbr describe how many cores are used for the radiation part.

TotalCores = (nproca * nprocb) + (nprocar * nprocbr)

 

2)

Three useful cases can be distinguished for the Concurrent Radiation version of ECHAM:

a)

For real asynchronous radiation behavior you want:

nprocar > 0 (e.g. = nproca)

nprocbr > 0 (e.g. = nprocb)

lrad_async = true

 

b)

Asynchronous radiation code is active in principle, but execution is still (forced) synchronous (e.g. used for testing purposes):

nprocar > 0 (e.g. = nproca)

nprocbr > 0 (e.g. = nprocb)

lrad_async = false

 

c)

Asynchronous radiation code is inactive, i.e. classical/original ECHAM behavior of the Concurrent Radiation version of ECHAM:

nprocar = 0

nprocbr = 0

lrad_async = don't care

 

Best,

 

      Kai

 

Von: Miguel ***@***.***> 
Gesendet: Mittwoch, 30. Juni 2021 13:06
An: esm-tools/esm_runscripts ***@***.***>
Cc: khimstedt ***@***.***>; Mention ***@***.***>
Betreff: Re: [esm-tools/esm_runscripts] fix/proc and core calc (#151)

 

a. Should nprocar/nprocbr be independent of nproca/nprocb at the time of calculating the number of tasks/cores or should we on the contrary only check for nprocar if we already have an nproca defined?

I changed my mind, I think they are independent and what @dbarbi <https://github.com/dbarbi>  did is probably the best:

1.	Check if nprocar/nprocbr have a valid value and define the number of tasks
2.	If not, try to find nproca/nprocb and define the number of tasks
3.	If not, try to find nproc and define the number of tasks.

The problem I was having can be solved by checking the values of nprocar/nprocbr to don't be either "remove_from_namelist" or 0.

Question b is still open.

mandresm:
Thanks @khimstedt , your comments solved all my questions.

esm_runscripts/batch_system.py
```diff
@@ -196,28 +196,31 @@ def calculate_requirements(config, cluster = None):
...
-                elif "nproca" in config[model] and "nprocb" in config[model]:
-                    config[model]["tasks"] = config[model]["nproca"] * config[model]["nprocb"]
+
+                if "nproca" in config[model] and "nprocb" in config[model]:
+                    config[model]["tasks"] = \
+                        config[model]["nproca"] * config[model]["nprocb"]
                     #end_proc = start_proc + int(config[model]["nproca"])*int(config[model]["nprocb"]) - 1
+                    # KH 30.04.20: nprocrad is replaced by more flexible
+                    # partitioning using nprocar and nprocbr
+                    no_nprocr = ["remove_from_namelist", 0]
```
pgierz: OK, after reading this twice, I understand what this is doing. Maybe put some comments? Or maybe a better name: `remove_nprocr_options` or something

pgierz (changes requested):
OK, looks correct, but I had a suggestion about the naming.

esm_runscripts/batch_system.py
mandresm: Thanks for the review :) , I just added the commits and I'll merge now to `prep_release`.

PR #152, opened 2021-07-01
Hotfix/better yaml dumper

denizural:
- Removed the topmost key `dictitem`, which is placed by `pyyaml` module. Thanks to the supreme documentation, it took few hours to figure out how to get it out. 
- Now there is no need to things like `config = read_config_file['dictitems']`
- No break up of long string on to multiple lines. Check any finished_config.yaml file. Many strings are broken into multiple lines although they are not in the incoming YAML files
- removed `strip_python_tags()` function and replaced it with added support to esm-tools data types, such as
    - `EsmToolsDir`
    - `Calendar`
    - ...
- Added support for `tuple`, which is missing in the original `pyyaml`
- By adding support for some data types, eg. `coupler_class`, the output is much cleaner. Eg. it does not dump the complete `namcouple` file. 
- The output is exactly the same apart from improved parts. Use `vimdiff` with `set diffopt+=iwhite` to compare
- Provides a good starting point for implementing support for other data types, in case they need to enter the `finished_config.yaml` file and need better formatting.

### DIFF
--- a/esm_runscripts/compute.py
+++ b/esm_runscripts/compute.py
@@ -309,7 +309,11 @@ def initialize_experiment_logfile(config):
 
 
 def _write_finalized_config(config):
-    """Writes <expid>_finished_config.yaml file"""
+    """Writes <expid>_finished_config.yaml file
+    Parameters
+    ----------
+    config : esm-tools config object
+    """
     # first define the representers for the non-built-in types, as recommended
     # here: https://pyyaml.org/wiki/PyYAMLDocumentation
     def date_representer(dumper, date):
@@ -317,7 +321,7 @@ def _write_finalized_config(config):
         
     def calendar_representer(dumper, calendar):
         # Calendar has a __str__ method
-        return dumper.represent_str(calendar.__str__())
+        return dumper.represent_str(str(calendar))
 
     def batch_system_representer(dumper, batch_system):
         return dumper.represent_str(f"{batch_system.name}")
@@ -365,7 +369,8 @@ def _write_finalized_config(config):
         config_final = copy.deepcopy(config) #PrevRunInfo
         del config_final["prev_run"]         #PrevRunInfo
 
-        out = yaml.dump(config_final, Dumper=EsmConfigDumper, width=10000)   #PrevRunInfo
+        out = yaml.dump(config_final, Dumper=EsmConfigDumper, width=10000, 
+            indent=4)   #PrevRunInfo
         config_file.write(out)
     return config
     
--- a/esm_runscripts/compute.py
+++ b/esm_runscripts/compute.py
@@ -4,13 +4,15 @@ import subprocess
 import copy
 import pathlib
 
-import esm_rcfile
 import six
 import yaml
-from esm_calendar import Date
 from colorama import Fore, Back, Style, init
 
 import esm_tools
+import esm_calendar
+import esm_parser
+import esm_rcfile
+import esm_runscripts
 
 from .batch_system import batch_system
 from .filelists import copy_files, log_used_files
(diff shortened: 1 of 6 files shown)


pgierz:
Looks good. Can you post a (brief/truncated) example yaml of what it looks like? Part of the reason I was trying to get rid of those tags is that I wanted to be able to read in the finished yaml again, but just using the standard yaml library. That would then complain: "I have read a tag for Calendar, but do not know how to do that"

esm_runscripts/compute.py
```diff
@@ -307,35 +309,66 @@ def initialize_experiment_logfile(config):
 
 
 def _write_finalized_config(config):
+    """Writes <expid>_finished_config.yaml file"""
+    # first define the representers for the non-built-in types, as recommended
+    # here: https://pyyaml.org/wiki/PyYAMLDocumentation
```
pgierz, line 318: I hate this documentation. With a passion. I'd ask you to add an anchor or something so we see where that is recommended, but it's just one HTML, without even clickable/jumpable section.

And it's always comforting to read:
> Reference
>
> Warning: API stability is not guaranteed!

esm_runscripts/compute.py
```diff
@@ -307,35 +309,66 @@ def initialize_experiment_logfile(config):
...
     def date_representer(dumper, date):
-        return dumper.represent_str("%s" % date.output())
+        return dumper.represent_str(f"{date.output()}")
+        
+    def calendar_representer(dumper, calendar):
+        # Calendar has a __str__ method
+        return dumper.represent_str(calendar.__str__())
```
pgierz: Why not `str(calendar)`?

mandresm:
>  Part of the reason I was trying to get rid of those tags is that I wanted to be able to read in the finished yaml again, but just using the standard yaml library.

In fact, that we can load finished yaml files again in the standard way is crucial for the `prev_run` feature. I might be able to test if it still works.

esm_runscripts/compute.py
```diff
@@ -307,35 +309,66 @@ def initialize_experiment_logfile(config):
...
+        # prevent dumping of whole namcouple
+        return dumper.represent_str(f"{coupler.name}")
+
+    def oasis_representer(dumper, oasis):
+        return dumper.represent_str(f"{oasis.name}")
 
-    yaml.add_representer(Date, date_representer)
-    yaml.add_representer(batch_system, batch_system_representer)
+    # dumper object for the ESM-Tools configuration
+    class EsmConfigDumper(yaml.dumper.Dumper):
```
pgierz, line 337: Nice idea, if we ever need to modify the dumper

pgierz (approved):

esm_runscripts/compute.py
denizural: I think that failed when I tried but let me check it again.

denizural: Ok, I tried again. Now it works. I was also puzzled when it did not but probably that was another thing causing the issue.

esm_runscripts/compute.py
denizural: Unfortunately pyyaml has a very 💩 documentation. I checked the HTML source. There are no anchors. I fixed the docstring anyways.

denizural:
@pgierz 
> Looks good. Can you post a (brief/truncated) example yaml of what it looks like? Part of the reason I was trying to get rid of those tags is that I wanted to be able to read in the finished yaml again, but just using the standard yaml library. That would then complain: "I have read a tag for Calendar, but do not know how to do that"

You can find examples of finished YAML outputs here:
`/work/ab0995/a271096/better_yaml_dumper`
- **test_finished_config_3.yaml:** 4 space indentation, tags removed. `[hotfix/better_yaml_dumper 35a1284]`
- **test_finished_config.yaml:** `44e17e8 (HEAD -> release, tag: v5.1.22, origin/release, origin/HEAD) Bump version: 5.1.21 → 5.1.22`

View the differences with:
`vimdiff test_finished_config_release.yaml test_finished_config_3.yaml`
and `:set diffopt+=iwhite` to ignore the whitespace

@mandresm When `prevrun` stuff comes, I can also implement a special dumping too.

mandresm:
> @pgierz
> 
> > Looks good. Can you post a (brief/truncated) example yaml of what it looks like? Part of the reason I was trying to get rid of those tags is that I wanted to be able to read in the finished yaml again, but just using the standard yaml library. That would then complain: "I have read a tag for Calendar, but do not know how to do that"
> 
> You can find examples of finished YAML outputs here:
> `/work/ab0995/a271096/better_yaml_dumper`
> 
>     * **test_finished_config_3.yaml:** 4 space indentation, tags removed. `[hotfix/better_yaml_dumper 35a1284]`
> 
>     * **test_finished_config.yaml:** `44e17e8 (HEAD -> release, tag: v5.1.22, origin/release, origin/HEAD) Bump version: 5.1.21 → 5.1.22`
> 
> 
> View the differences with:
> `vimdiff test_finished_config_release.yaml test_finished_config_3.yaml`
> and `:set diffopt+=iwhite` to ignore the whitespace
> 
> @mandresm When `prevrun` stuff comes, I can also implement a special dumping too.

`prev_run` stuff is already in `release` since a couple of weeks. I will test if everything still works and report back.

mandresm:
Restarts and branching off experiments tested for `AWICM3` with a fix to still support old config files containing `dictitems`.

mandresm (approved):

Issue #156, opened 2021-07-07
Not copying the script to the EXP_PATH/scripts


mandresm:
@xshi-awi found a bug in `esm_runscripts` consisting in the following:

Trying to run `esm_runscripts <script.yaml> -e <exp_id>` from the `base_dir` (e.g. in mistral for me that is `/work/ab0246/a270152/esm_yaml_test/`) results in the `<script.yaml>` not being copied into the `<EXP_PATH>/scritps`. This of course breaks the whole simulation because ESM-Tools relays on readying the script from the experiment directory. If you try to run your runscript from a folder has some uncommon parents with the `scriptsdir` then everything works as it should.

I think the reason for that is one of the lines of the last patches in release:
https://github.com/esm-tools/esm_runscripts/blob/44e17e88a110507174461da435b63155725ab05e/esm_runscripts/compute.py#L508

If `from_dir` is in my case `/mnt/lustre02/work/ab0246/a270152/esm_yaml_test/`, cause my runscript is there, and `scriptsdir` is `/mnt/lustre02/work/ab0246/a270152/esm_yaml_test/xiaoxu/scripts`, then `scriptsdir_inside_fromdir` is gonna be true, i.e. `for p in pathlib.Path(scriptsdir).parents: p` returns the following:
```
PosixPath('/mnt/lustre02/work/ab0246/a270152/esm_yaml_test/xiaoxu')
PosixPath('/mnt/lustre02/work/ab0246/a270152/esm_yaml_test')
PosixPath('/mnt/lustre02/work/ab0246/a270152')
PosixPath('/mnt/lustre02/work/ab0246')
PosixPath('/mnt/lustre02/work')
PosixPath('/mnt/lustre02')
PosixPath('/mnt')
PosixPath('/')
```

As esm_runscripts is designed to run the runscript from the scripts directory my suggested solution is that we modify that line to be:
```
scriptsdir_inside_fromdir = pathlib.Path(fromdir) == pathlib.Path(scriptsdir)
```
What do you think @denizural , @dbarbi ?

**Versions used**
```
+---------------------+-----------+---------------------------------------------------------------+----------+-------------------+
| package_name        | version   | file                                                          | branch   | tags              |
|---------------------+-----------+---------------------------------------------------------------+----------+-------------------|
| esm_calendar        | 5.0.0     | /mnt/lustre01/pf/a/a270152/.local/lib/python3.6/site-packages |          |                   |
[...]

PR #157, opened 2021-07-08
replaced inside flag with deep nesting

denizural:
This solves the issue https://github.com/esm-tools/esm_runscripts/issues/156

I was able to replicate the issue that @mandresm reported. It turns out to be that if the runscript is inside the workdir (where model also runs) and `--update` is on the anti-recursion guard is activated (friendly fire). That guard needs to be there but we clearly need a solution.

`scriptsdir_inside_fromdir` is replaced by `deep_nesting_found`. It basically goes 6 directories up and checks if the `scriptsdir` is inside that directory and its parents. If some kind of infinite recursion occurs (should not occur but prepare for the worst), it will not proceed infinitely. Those infinite recursion was encountered in some experimental runs from Paul, and probably due to the newly implemented `!ENV` feature at that time.

I checked the following cases:
- release branch, no update, workdir
  - **result:** no YAML runscript (or secondary script, eg. `further_reading`) in the `expid/scripts` directory.

- bugfix branch, no update, workdir
  - **result:** the missing files are there 

**Comparison:**
- Enter each `expid` and 
```bash
find . &> ../file_list_release
find . &> ../file_list_bugfix
```
- Then vimdiff: `vimdiff file_list_release file_list_bugfix`

The only differences were the YAML files in the `expid/scripts` directory

### DIFF
--- a/esm_runscripts/compute.py
+++ b/esm_runscripts/compute.py
@@ -471,6 +471,8 @@ def copy_tools_to_thisrun(config):
     fromdir = os.path.realpath(gconfig["started_from"])
     # Directory where the runscript is copied to
     scriptsdir = os.path.realpath(gconfig["experiment_scripts_dir"])
+    # basedir + expid
+    expdir = os.path.realpath(gconfig['experiment_dir'])
 
     # Paths inside the experiment directory where esm_tools and namelists
     # are copied to. Those are not functional but a reference to what was
@@ -505,16 +507,24 @@ def copy_tools_to_thisrun(config):
 
     # check for recursive creation of the file tree. This prevents the risk of
     # creating a run directory tree inside the `scriptsdir`
-    scriptsdir_inside_fromdir = pathlib.Path(fromdir) in pathlib.Path(scriptsdir).parents
-    if scriptsdir_inside_fromdir and config["general"]["verbose"]:
+    # example:
+    # fromdir:    /mnt/lustre02/work/project/username/basedir
+    # scriptsdir: /mnt/lustre02/work/project/username/basedir/expid/scripts
+    # it is assumed that if 5th parent of the scriptsdir is inside the fromdir
+    # then probably some sort of infinite recursion is entered. These lines
+    # protect such problems
+    scriptsdir_deep_parents = list(pathlib.Path(scriptsdir).parents)[5:]
+    deep_nesting_found = pathlib.Path(expdir) in scriptsdir_deep_parents
+    if deep_nesting_found and config["general"]["verbose"]:
         print("WARNING: scriptsdir is inside fromdir")
-        print(f"  - scriptsdir: {scriptsdir}")
-        print(f"  - fromdir:    {fromdir}")
+        print(f"  - scriptsdir:         {scriptsdir}")
+        print(f"  - fromdir:            {fromdir}")
+        print(f"  - experiment dir:     {expdir}")
     
     # If ``fromdir`` and ``scriptsdir`` are the same, this is already a computing
     # simulation which means we want to use the script in the experiment folder,
     # so no copying is needed
-    if (fromdir == scriptsdir) or scriptsdir_inside_fromdir and not gconfig["update"]:
+    if (fromdir == scriptsdir) or deep_nesting_found and not gconfig["update"]:
         if config["general"]["verbose"]:
             print("Started from the experiment folder, continuing...")
         return config
--- a/esm_runscripts/compute.py
+++ b/esm_runscripts/compute.py
@@ -11,6 +11,7 @@ from esm_calendar import Date
 from colorama import Fore, Back, Style, init
 
 import esm_tools
+import esm_parser
 
 from .batch_system import batch_system
 from .filelists import copy_files, log_used_files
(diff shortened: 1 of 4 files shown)


esm_runscripts/compute.py
```diff
@@ -505,16 +507,24 @@ def copy_tools_to_thisrun(config):
...
-    if scriptsdir_inside_fromdir and config["general"]["verbose"]:
+    # example:
+    # fromdir:    /mnt/lustre02/work/project/username/basedir
+    # scriptsdir: /mnt/lustre02/work/project/username/basedir/expid/scripts
+    # it is assumed that if 5th parent of the scriptsdir is inside the fromdir
+    # then probably some sort of infinite recursion is entered. These lines
+    # protect such problems
+    scriptsdir_deep_parents = list(pathlib.Path(scriptsdir).parents)[5:]
+    deep_nesting_found = pathlib.Path(expdir) in scriptsdir_deep_parents
+    if deep_nesting_found and config["general"]["verbose"]:
```
mandresm: Isn't `deep_nesting_found = pathlib.Path(expdir) in scriptsdir_deep_parents` always gonna be `true`? I mean `expdir` (`<EXP_DIR>`) is always gonna be parent of `scriptsdir` (`<EXP_DIR>/scripts`) right?

Also I still don't understand this new approach, the question we are trying to answer here is "is the script given in the user command the one inside the scripts folder in the experiment?" right?

In that case, I don't understand why we need to check for nesting paths and parents. I know it is related to the infinite loop issue you mentioned before.

But let's say for example you have your `scriptsdir_deep_parents` be (that's the case for the ALEPH computer):
```
PosixPath('/scratch/awi/mandresm/test/scripts') -> 0
PosixPath('/scratch/awi/mandresm/test') -> 1
PosixPath('/scratch/awi/mandresm') -> 2
PosixPath('/scratch/awi') -> 3
PosixPath('/scratch') -> 4
PosixPath('/') -> 5
```

Then the problem would still remain right?

mandresm (changes requested):

esm_runscripts/compute.py
denizural: Hi @mandresm,

I don't know if I get this correctly but here is an analogy:

You can think of 5 here as a _maximum iteration count_ in numerical models. If any childhood is met after the 5th level it means that something strange is going on since the runscript is copied to a very deep location. But in the normal case this does not happen. Eg. convergence is met before the max iteration number. 

Here is a code snippet that you can try in IPython:

```python
import pathlib

# scenario 1: normal case, no recursion 
expdir = pathlib.Path("/mnt/lustre02/work/ab0995/a271096/runscripts/issue_156")
scrdir = pathlib.Path("/mnt/lustre02/work/ab0995/a271096/runscripts/issue_156/scripts/")

print(expdir in list(scrdir.parents))        # True
print(expdir in list(scrdir.parents)[5:])    # False

# scenario 2: recursive error is met for some crazy reason
scrdir = pathlib.Path("/mnt/lustre02/work/ab0995/a271096/runscripts/issue_156/scripts/runscripts/issue_156/scripts/runscripts/issue_156/scripts/")
print(expdir in list(scrdir.parents))       # True
print(expdir in list(scrdir.parents)[5:])   # True

# It becomes True and guard is activated because

list(scrdir.parents)[5:]
[PosixPath('/mnt/lustre02/work/ab0995/a271096/runscripts/issue_156/scripts'),
 PosixPath('/mnt/lustre02/work/ab0995/a271096/runscripts/issue_156'),
 PosixPath('/mnt/lustre02/work/ab0995/a271096/runscripts'),
 PosixPath('/mnt/lustre02/work/ab0995/a271096'),
 PosixPath('/mnt/lustre02/work/ab0995'),
 PosixPath('/mnt/lustre02/work'),
 PosixPath('/mnt/lustre02'),
 PosixPath('/mnt'),
 PosixPath('/')]
 
print(fromdir)
PosixPath('/mnt/lustre02/work/ab0995/a271096/runscripts')
```

I had to implement this check because the infinite recursion is really nasty. It creates multiple instances of `esm_runscripts` and you need to use `killall` command to kill them all since it will keep spawning new instances 😱 , meanwhile you think that your simulation is running.

mandresm: Thanks for the clarification

denizural: After the discussion, I realized that this fix can be done a better way. So, I removed the `deep_nesting_found` in the if statement and added a call to `esm_parser.user_error()` to exit elegantly. 

It is still important to note that when deep recursive occurs there will be multiple instances of `esm_runscripts` running in the background. Therefore, maybe a further call to `os.kill()` can be added. But I think it is Ok for now. As `esm_runscripts` become more robust, we won't run into that recursion issue.

mandresm (approved):

esm_runscripts/compute.py
```diff
@@ -505,16 +508,31 @@ def copy_tools_to_thisrun(config):
...
-        print("WARNING: scriptsdir is inside fromdir")
-        print(f"  - scriptsdir: {scriptsdir}")
-        print(f"  - fromdir:    {fromdir}")
+    # example:
+    # fromdir:    /mnt/lustre02/work/project/username/basedir
+    # scriptsdir: /mnt/lustre02/work/project/username/basedir/expid/scripts
+    # it is assumed that if 5th parent of the scriptsdir is inside the fromdir
+    # then probably some sort of infinite recursion is entered. These lines
+    # protect such problems
+    scriptsdir_deep_parents = list(pathlib.Path(scriptsdir).parents)[5:]
```
pgierz, line 517: I don't really like the hardcoded 5 here....this is **very** specific to how mistral's filesystem is set up.

denizural: 5 is more like a threshold for detecting infinite recursion. Typically `esm_tools` directory tree is not very deep. This is not really about mistral since we start counting 5 from the deepest level (right to left). It is machine agnostic.

PR #171, opened 2021-08-03
Feat/transient forcing

pgierz:
Allows for the user to set transient forcing for a ECHAM run.

@dbarbi, is this what Uta wanted?


dbarbi:
yup, i guess it is

pgierz:
OK. Can you get in touch with her and ask her to check if it works? My own test went through, but it might be good to get a second opinion. There might be some minor things not working 100%, you never know....

PR #173, opened 2021-08-10
Prep release

ackerlar:
added fix to avoid that precompute or compute start before necessary files are in place from coupling shell scripts


ackerlar:
@pgierz I added the loop and it seems to work fine. You have to specify in `general` something like:
```
        files_to_wait_for: 
            - "/work/ba0989/a270124/125ka_FINAL_TEST03/couple/latest_atmo_forcing_file.nc"
            - "/work/ba0989/a270124/125ka_FINAL_TEST03/couple/latest_ocean_forcing_file.nc"
```

esm_runscripts/namelists.py
```diff
@@ -305,7 +306,37 @@ def nmls_output_all(config):
...
+    @staticmethod
+    def apply_iceberg_calving(config):
+        """
+        Applies a disturbance to the DYNCTL chapter of the echam namelist via the enstdif
 
+        Relevant configuration entries:
+        * disturbance_years (list of int): Which year to apply the disturbance
+        * distrubance (float): Value to apply. Default can be found in echam.yaml
+        """
+        if "fesom" in config["general"]["valid_model_names"] and config["fesom"]["use_icebergs"] and config["fesom"]["use_icesheet_coupling"]:
```
pgierz: To avoid key errors, I would switch this to `get` instead

esm_runscripts/config_initialization.py
```diff
@@ -14,7 +14,7 @@ def init_first_user_config(command_line_config, user_config):
     # maybe switch to another runscript, if iterative coupling
     if user_config["general"].get("iterative_coupling", False):
         user_config = chunky_parts.setup_correct_chunk_config(user_config)
-        next_model = user_config["general"]["original_config"]["general"]["model_queue"][1]
+        next_model = user_config["general"]["original_config"]["general"]["model_queue"][-1]
```
pgierz: don't really understand this diff... shouldn't you just be rotating to the "next" model, instead of taking the last one?

esm_runscripts/coupler.py
```diff
@@ -16,6 +16,9 @@ def __init__(self, full_config, name):
         if "coupling_target_fields" in full_config[self.name]:
             for restart_file in list(full_config[self.name]["coupling_target_fields"]):
                 self.nb_of_couplings += len(list(full_config[self.name]["coupling_target_fields"][restart_file]))
+        if "coupling_input_fields" in full_config[self.name]:
```
pgierz, line 19: this one is for oasis fields marked as INPUT? Or what exactly?

esm_runscripts/namelists.py
```diff
@@ -305,7 +306,37 @@ def nmls_output_all(config):
         for model in config["general"]["valid_model_names"]:
             config[model] = nmls_output(config[model], config["general"]["verbose"])
         return config
+    
+    @staticmethod
+    def apply_iceberg_calving(config):
```
pgierz, line 311: Hooray for doc-strings. We need them everywhere. Thanks Lars :-)

pgierz (changes requested):

pgierz:
Requested changes: make sure to use `get` instead of direct access with `[key-name]`, to avoid a key error.

esm_runscripts/config_initialization.py
ackerlar: Yes, I did this when I experimented with two PISM instances, i.e. in total three models (awiesm, pism1, pism2) and it worked out better. But I guess it's better to leave it at the old value at the moment.

esm_runscripts/coupler.py
ackerlar: Yes, it was necessary to include the lgfw field into namcouple

esm_runscripts/config_initialization.py
pgierz: What does `model queue` look like? is that a list? Because at some point (when we have vilma, pism, and climate all at once) we will have three models. So here we need a way to rotate around each model, something like:
```python
# Assume you have
model_queue = ["awicm", "pism", "vilma"]
current_model = user_config['general']['original_config']['general']['model_queue'].pop()
# Now model queue is something like ["pism", "vilma"] and current model is "awicm"
user_config['general']['original_config']['general']['model_queue'].append(current_model)
# Now model queue is:
model_queue = ["pism", "vilma", "awicm"]
```

in all likelihood there is a python build-in function to do this, but that is how I would solve it. I need more context though to know that is what you actually want to do here...

dbarbi: I had to change the same lines for wiso, and pushed that as a bugfix already. Please check. I think [-1] doesn't do what we want.

ackerlar: I have changed it back already.

mandresm:
Can this be then merged?

pgierz:
yes

pgierz (approved):

Issue #172, opened 2021-08-03
esm_runscript crash when scenario is added


seb-wahl:
With the latest version (on HLRN4):
```
glogin5:~/esm/esm_tools/runscripts/foci $ esm_versions check
+---------------------+-----------+---------------------------------------------------+----------+----------------------------+
| package_name        | version   | file                                              | branch   | tags                       |
|---------------------+-----------+---------------------------------------------------+----------+----------------------------|
| esm_calendar        | 5.0.0     | /home/shkifmsw/.local/lib/python3.7/site-packages |          |                            |
| esm_database        | 5.0.0     | /home/shkifmsw/.local/lib/python3.7/site-packages |          |                            |
| esm_environment     | 5.1.1     | /home/shkifmsw/.local/lib/python3.7/site-packages |          |                            |
| esm_master          | 5.1.3     | /home/shkifmsw/.local/lib/python3.7/site-packages |          |                            |
| esm_motd            | 5.0.2     | /home/shkifmsw/.local/lib/python3.7/site-packages |          |                            |
| esm_parser          | 5.1.7     | /home/shkifmsw/.local/lib/python3.7/site-packages |          |                            |
| esm_plugin_manager  | 5.0.1     | /home/shkifmsw/.local/lib/python3.7/site-packages |          |                            |
| esm_profile         | 5.0.0     | /home/shkifmsw/.local/lib/python3.7/site-packages |          |                            |
| esm_rcfile          | 5.1.0     | /home/shkifmsw/.local/lib/python3.7/site-packages |          |                            |
| esm_runscripts      | 5.1.27    | /home/shkifmsw/esm/esm_runscripts                 | release  | v5.1.27-2-g9405139         |
| esm_tools           | 5.0.12    | /home/shkifmsw/esm/esm_tools                      | geomar   | v5.0.12-161-ge989858-dirty |
| esm_version_checker | 5.1.5     | /home/shkifmsw/.local/lib/python3.7/site-packages |          |                            |
+---------------------+-----------+---------------------------------------------------+----------+----------------------------+
```
I added 
```
echam:
   scenario: 1percCO2
```
(I want to prepare a 1pct CO2 scenario for a colleague) to our FOCI default setup. If I add the `scenario: 1percCO2` to the ECHAM6 block (note: If I add it to the `general: ` section, nothing happens), esm_runscripts crashed with the following:
```
[...]

PR #174, opened 2021-08-13
prints an explanatory error message if a scenario is not implemented …

denizural:
This PR solves the issue https://github.com/esm-tools/esm_runscripts/issues/172

I tested different solutions and finally decided that this one gives the most useful error message to the user. This error message provides the user
- which **model** & **version**,
- which **scenario** and
- which **file type** is causing the problem
- together with the **file source** causing the crash

eg. if you try to run ECHAM with the scenario ` scenario: 1percCO2`, now these lines will be printed:

```
ERROR: Missing Scenario Configuration
-------------------------------------

Scenario 1percCO2 for the model jsbach (version: 3.20p1) has not been implemented yet.
The input file variable LU_HIST of forcing_sources can not be fully resolved:

/work/bb0519/foci_input2/JSBACH/input/r0010/T63/New_Hampshire_LCC/hist_harvest/LUH_harvest_T63_1850.nc:
    to: 1850
/work/bb0519/foci_input2/JSBACH/input/r0010/T63/New_Hampshire_LCC/hist_harvest/LUH_harvest_T63_@YEAR@.nc:
    from: 1851
```

Instead of 
```
Traceback (most recent call last):
  File "/home/shkifmsw/.local/bin/esm_runscripts", line 11, in <module>
  File "/home/shkifmsw/esm/esm_runscripts/esm_runscripts/cli.py", line 202, in main
    Setup()
  File "/home/shkifmsw/esm/esm_runscripts/esm_runscripts/sim_objects.py", line 50, in __call__
    self.compute(*args, **kwargs)
  File "/home/shkifmsw/esm/esm_runscripts/esm_runscripts/sim_objects.py", line 84, in compute
    self.config = compute.run_job(self.config)
  File "/home/shkifmsw/esm/esm_runscripts/esm_runscripts/compute.py", line 32, in run_job
    config = evaluate(config, "compute", "compute_recipe")
  File "/home/shkifmsw/esm/esm_runscripts/esm_runscripts/helpers.py", line 49, in evaluate
    framework_recipe, framework_plugins, config
  File "/home/shkifmsw/.local/lib/python3.7/site-packages/esm_plugin_manager/esm_plugin_manager.py", line 138, in work_through_recipe
    config = getattr(submodule, workitem)(config)
  File "/home/shkifmsw/esm/esm_runscripts/esm_runscripts/filelists.py", line 941, in assemble
    config = complete_targets(config)
  File "/home/shkifmsw/esm/esm_runscripts/esm_runscripts/filelists.py", line 126, in complete_targets
    config[model][filetype + "_sources"][categ]
  File "/sw/tools/python/anaconda3/2019.03/skl/lib/python3.7/posixpath.py", line 146, in basename
    p = os.fspath(p)
TypeError: expected str, bytes or os.PathLike object, not dict
```

I tried the same hack in other parts of the `esm_runscripts` but did not find anything yet. Maybe someone else will in the future and then we can use the same solution there too.

After the merge and bump, I will merge the `HIST` case stuff.

### DIFF
--- a/esm_runscripts/filelists.py
+++ b/esm_runscripts/filelists.py
@@ -12,6 +12,7 @@ import esm_parser
 
 import f90nml
 import esm_tools
+import yaml
 
 def rename_sources_to_targets(config):
     # Purpose of this routine is to make sure that filetype_sources and
@@ -130,9 +131,23 @@ def complete_targets(config):
             if filetype + "_sources" in config[model]:
                 for categ in config[model][filetype + "_sources"]:
                     if not categ in config[model][filetype + "_targets"]:
-                        config[model][filetype + "_targets"][categ] = os.path.basename(
-                            config[model][filetype + "_sources"][categ]
-                        )
+                        file_source = config[model][filetype + "_sources"][categ]
+
+                        # check if the file_source has the correct type. For
+                        # unresolved variables they may still be a 'dict'
+                        if not isinstance(file_source, (str, os.PathLike)):
+                            scenario = config[model].get('scenario', 'UNDEFINED')
+                            version = config[model].get('version', 'UNDEFINED')
+                            error_type = "Missing Scenario Configuration"
+                            error_text = (
+                                f"Scenario {scenario} for the model {model} (version: {version}) has not been implemented yet. \n" +
+                                f"The input file variable {categ} of {filetype}_sources can not be fully resolved:\n\n" +
+                                yaml.dump(file_source, indent=4))
+                            esm_parser.user_error(error_type, error_text)
+                        else:
+                            config[model][filetype + "_targets"][categ] = \
+                                os.path.basename(file_source)
+
     return config
 
 
--- a/esm_runscripts/filelists.py
+++ b/esm_runscripts/filelists.py
@@ -141,7 +141,7 @@ def complete_targets(config):
                             scenario = config[model].get('scenario', 'UNDEFINED')
                             version = config[model].get('version', 'UNDEFINED')
 
-                            error_type = "Missing Scenario Configuration"
+                            error_type = "Missing Configuration"
                             error_text = (
                                 # comment-out the line below to provide more information
                                 # f"Scenario {scenario} for the model {model} (version: {version}) has not been implemented yet. \n" +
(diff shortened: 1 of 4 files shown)


mandresm:
This is very specific to models where `scenario` plays role and `_sources` are a dictionary. It could be that in the future we add more functionalities to the `_sources` making it sometimes a `dict`, or that other models use a different name than `scenario` for their case variable. Then the error message will be still confusing. I'm not sure, it looks that it needs to be a bit more general. What about not mentioning the scenario but just that the error looks like something like this:

```
ERROR: Sources not properly resolved
---------------------------------------

The input file variable LU_HIST of forcing_sources can not be fully resolved:

/work/bb0519/foci_input2/JSBACH/input/r0010/T63/New_Hampshire_LCC/hist_harvest/LUH_harvest_T63_1850.nc:
    to: 1850
/work/bb0519/foci_input2/JSBACH/input/r0010/T63/New_Hampshire_LCC/hist_harvest/LUH_harvest_T63_@YEAR@.nc:
    from: 1851
```

However, if `_sources` being a dictionary always depends on `scenario` and there is a specific handling of the `scenario` variable inside `esm_runscripts` then the fix is already good, as it would only happen when `scenario` is in use.

seb-wahl:
Although I like the very precise error message, I think it's too specific. I would add the model version to the error message and leave out the scenario specific stuff. I vaguely think to remember that I had this error message in other cases, too, when @YEAR@ things could not be resolved. But apart from that. Thanks for the quick and solid fix to get rid of the confusing error message.

esm_runscripts/filelists.py
```diff
@@ -136,11 +136,15 @@ def complete_targets(config):
                         # check if the file_source has the correct type. For
                         # unresolved variables they may still be a 'dict'
                         if not isinstance(file_source, (str, os.PathLike)):
+                            # model, scenario and version are omitted to make
+                            # error message less verbose
                             scenario = config[model].get('scenario', 'UNDEFINED')
                             version = config[model].get('version', 'UNDEFINED')
+
                             error_type = "Missing Scenario Configuration"
```
seb-wahl, line 144: The error type needs to be renamed if we remove the Scenario details above. Suggestion: Missing Configuration i.e. remove Scenario from the text.

seb-wahl (changes requested):
The text for the error_type has to be adjusted. Suggestion: Missing configuration i.e. remove Scenario from the text

PR #176, opened 2021-08-16
Hist fixed

denizural:
support for HIST scenario in Echam. This should also get merged to release. It does not change the basic functionality we have so far but adds support for file sources that have `from` and `to` variables. 

The solution `@YEAR_1850@` may not be the most optimal one since @seb-wahl seem to have an easier solution. I can remove that later on.


seb-wahl (approved):
I can confirm that this now works for our historical setups

seb-wahl:
Could this also go into release soon, or is it changes that you think are too large?

denizural:
Thank you very much Sebastian. 👍  I was waiting for your approval. Yes, I will push that to develop. Perform some tests comparing to my previous hist runs and make another PR to the release.

mandresm:
Tested with the following models:
- AWICM2.0
- AWIESM2.1-REcoM
- FESOM2.0/2.1

I'll merge it now.

PR #178, opened 2021-08-18
debug_level (NLOGPRT) can now be configured for the namcouple file fo…

joakimkjellsson:
The NLOGPRT (debug_level) was always set to "1" for OASIS. 
This little fix allows the user to set this in the runscript. 
For example, 
```bash
oasis3mct:
         debug_level: 50 
````
will force OASIS to print all fluxes before and after conservation (e.g. GLBPOS) which is very helpful when debugging a coupled model. 

Note that the Lucia tool must be turned off for this to work.

### DIFF
--- a/esm_runscripts/__init__.py
+++ b/esm_runscripts/__init__.py
@@ -2,7 +2,7 @@
 
 __author__ = """Dirk Barbi"""
 __email__ = 'dirk.barbi@awi.de'
-__version__ = "5.1.29"
+__version__ = "5.1.30"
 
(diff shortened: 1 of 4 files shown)


denizural:
I was also thinking of implementing that while I was debugging AWICM-1. Thanks @joakimkjellsson 😃

pgierz (approved):
Looks good. 

One thing we forgot to discuss yesterday: we will at some point need to (briefly) freeze everything so I can put together our "monorepo"

PR #186, opened 2021-09-14
hotfix/generalization of last minute

mandresm:
Generalization of last minute changes so that the file for last minute changes can now also contain standard 'runsript-looking-like' variables. This function is accessed by `esm_master` through the `-m <file.yaml>` flag through a call to `esm_runscripts.SimulationSetup` (https://github.com/esm-tools/esm_master/blob/39e8ae5f741ae36a861372ca0baad6ae5a90dd9b/esm_master/esm_master.py#L52-L54). This option is particularly useful for `esm_master` as it could allow the user to switch for example between MPI configurations, running `esm_master install-<model> -m <file.yaml>` where the file contains:
```
computer:
    useMPI: openmpi
```

**TODO**

- [ ] Document the feature `last_minute_changes` feature
- [x] Test `esm_master`
- [x] Test `esm_runscripts`

### DIFF
--- a/esm_runscripts/__init__.py
+++ b/esm_runscripts/__init__.py
@@ -2,7 +2,7 @@
 
 __author__ = """Dirk Barbi"""
 __email__ = 'dirk.barbi@awi.de'
-__version__ = "5.1.30"
+__version__ = "5.1.31"
 
 from .sim_objects import *
 from .batch_system import *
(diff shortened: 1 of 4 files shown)


pgierz:
HI Miguel, could you also maybe comment where the `-m <file>` gets put in? At the moment I only see dictionary merging in the PR, but (for example) I did not even know that this was possible to change . That for me points to a gap in the documentation...

mandresm:
@pgierz , I am not sure I fully understand the question. If what you are asking is how the info from the file gets into the `dict_merge` of the PR, that's something that happens through the instancing of the `last_minute` class:
https://github.com/esm-tools/esm_runscripts/pull/186/files#diff-15c3932681c3e75667b90e4247313f19b795c7bf07df4695d94de491f4f56e2fL5-L25

Does this answers your question? I agree, we need to document `-m` flag.

PR #188, opened 2021-10-06
fix for do_hosing

niceniulu:
changes for hosing

### DIFF
--- a/esm_runscripts/coupler.py
+++ b/esm_runscripts/coupler.py
@@ -16,6 +16,8 @@ class coupler_class:
         if "coupling_target_fields" in full_config[self.name]:
             for restart_file in list(full_config[self.name]["coupling_target_fields"]):
                 self.nb_of_couplings += len(list(full_config[self.name]["coupling_target_fields"][restart_file]))
+        if full_config['fesom'].get("do_hosing"):
+                self.nb_of_couplings += 1
         if name == "oasis3mct":
             from . import oasis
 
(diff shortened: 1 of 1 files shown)


esm_runscripts/coupler.py
```diff
@@ -16,14 +16,16 @@ def __init__(self, full_config, name):
         if "coupling_target_fields" in full_config[self.name]:
             for restart_file in list(full_config[self.name]["coupling_target_fields"]):
                 self.nb_of_couplings += len(list(full_config[self.name]["coupling_target_fields"][restart_file]))
+        if full_config['fesom'].get("do_hosing"):
```
mandresm: I think this line can be reached from a coupled setup that has no `fesom` component and therefore, would lead to a crash. I would suggest to change this line to something like:
```
if full_config.get('fesom', {}).get("do_hosing"):
```
This is in principle a more general and safe solution.

mandresm (changes requested):

esm_runscripts/coupler.py
mandresm:
```suggestion
        if full_config.get('fesom', {}).get("do_hosing"):
```

mandresm:
@pgierz, this was never implemented in `esm_runscripts` for release 6 (it's a pending PR in the old `esm-tools/esm_runscripts` package). Should it go in in the monorepo?

pgierz:
This could go in if we have a test runscript and expected result. 

Nice that you are going through the old packages @mandresm! We should see about erasing them soon (maybe in autumn?) I put most of the old things into "archive mode", so no one can push. You can still clone though, which means people might still end up with an old system.

denizural:
I think they should stay as read-only until we completely clean up the current repo. Then we can announce that old version is completely retired and erased. Of course, we can still host it somewhere safe. Eg. AWI GitLab

PR #193, opened 2021-10-21
Prep release

ackerlar:
- I have got a key error for `batch_or_shell` in `subjob_config`. The fix to esm_runscripts/dataprocess.py works for me
 - The last run in each chunk of the model2 (in a coupled setup) had a wrong chunk_number associated to it. The fix in esm_runscripts/chunky_parts.py works for me. But it is only tested for two coupled models

### DIFF
--- a/esm_runscripts/chunky_parts.py
+++ b/esm_runscripts/chunky_parts.py
@@ -98,6 +98,10 @@ def set_chunk_calendar(config):
     next_date = config["general"]["next_date"]
 
     number_of_chunks_done = this_chunk_number // number_of_models
+   
+    # LA: get right number_of_chunks_done for last year in chunk
+    if config["general"]["model_queue"][0] == "model1" and config["general"]["last_run_in_chunk"]:
+        number_of_chunks_done = number_of_chunks_done - 1
     
     passed_time = ( 
             number_of_chunks_done * chunk_delta_date[0],
--- a/esm_runscripts/batch_system.py
+++ b/esm_runscripts/batch_system.py
@@ -123,20 +123,60 @@ class batch_system:
                 ("@jobtype@", cluster),
                 ]
 
-        all_flags = [
-            "partition_flag",
-            "time_flag",
-            tasks_nodes_flag,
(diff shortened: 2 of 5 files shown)


esm_runscripts/chunky_parts.py
```diff
@@ -98,6 +98,10 @@ def set_chunk_calendar(config):
     next_date = config["general"]["next_date"]
 
     number_of_chunks_done = this_chunk_number // number_of_models
+   
+    # LA: get right number_of_chunks_done for last year in chunk
+    if config["general"]["model_queue"][0] == "model1" and config["general"]["last_run_in_chunk"]:
+        number_of_chunks_done = number_of_chunks_done - 1
```
mandresm, line 104: Why do we need to substract 1 to the `number_of_chunks_done` only for `model1` and not for `model2` (or also `model3` 4, 5, ...)?

ackerlar: It is because of this:
```

    number_of_models = config["general"]["number_of_ic_models"]
    this_chunk_number = int(config["general"]["next_chunk_number"]) - 1
    ...

    number_of_chunks_done = this_chunk_number // number_of_models
```

When you, for example, run AWI-ESM and PISM in 5 year chunks each. And you are within in first PISM chunk (which is the second chunk in total):
`number_of_models` = 2
`next_chunk_number` = 2 (for years 1 to 4) and 3 (for last year in this chunk)
then: 
`this_chunk_number` = 1 (for years 1 to 4) and 2 (for last year in chunk)
`number_of_chunks_done` = 0 (for years 1 to 4) and 1 (for last year in chunk, although it is still the first PISM chunk)

So I guess it would be necessary to it for every second model. It is a bit unlucky that it says: 
`if config["general"]["model_queue"][0] == "model1"` while it is actually applied to `model2`. Maybe one can also say::
`if config["general"]["model_queue"][1] == "model2"` 
I will test this.

mandresm:
This is moved to `esm-tools/esm_tools#547`, and therefore, closed here.

PR #202, opened 2021-11-08
hotfix/namelist remove var properties

mandresm:
`remove_from_namelist` can now remove variables with properties such as `YS_NL%LMASSFIX`


mandresm:
I'll soon as this is approve I will add this to the monorepo.

PR #203, opened 2021-11-09
Hotfix/namelist remove var properties

mandresm:
Same as #202

### DIFF
--- a/esm_runscripts/__init__.py
+++ b/esm_runscripts/__init__.py
@@ -2,7 +2,7 @@
 
 __author__ = """Dirk Barbi"""
 __email__ = 'dirk.barbi@awi.de'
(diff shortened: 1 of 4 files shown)


pgierz:
Approve, please put in monorepo also.

Sorry, I am only halfway there today, I  am at a data science workshop. Best quote so far has been:

> we have too many flavors of filenames coming from the sensor. And then, when we built the app, someone tried to just name their filenames “A.csv”, “B.csv”, “C.csv”, and the scientist who worked on this retired, so we were unable to understand what was going on.

So, by comparison, I would say esm-tools is only minimally confusing ;-)

mandresm:
Now, also in the monorepo

